PR #10, opened 2016-07-17
Replace word “beers” for “drinks”

lukemorton:
Couple reasons:

 - I’m drinking more Gin and Tonics these days
 - Some people prefer a Coca-Cola
 - It’s more about socialising than a particular beverage

In the future I’d even prefer to frame the evening around the word social rather than drinks. It’s all about inclusion of various view points and lets be fair some people don’t do alcohol.


davidwinter:
LGTM!

PR #15, opened 2016-08-16
Remove +Support and +AaaS traits

lukemorton:
We are removing these traits as we no longer see a separation between Support and AaaS work. Engineers should be able to move freely between our various offerings.

### DIFF
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -12,7 +12,7 @@ We need someone:
 
 We expect someone in the role:
 
-- to be billable for at least half their time after 2 weeks on-boarding
+- to be fully billable after 2 weeks on-boarding
 - to be involved in practice improvement discussions within 3 months
 - to be actively practicing modern software craftsmanship practices such as TDD within 3 months
 - to be autonomously delivering features of a small nature within 3 months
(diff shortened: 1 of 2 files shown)


davidwinter:
LGTM!

roles/engineer.md
```diff
@@ -12,37 +12,14 @@ We need someone:
 
 We expect someone in the role:
 
+- to be billable for at least half their time after 2 weeks on-boarding
```
chrisblackburn: Historically the half-time billable expectations was for less experienced joiner
In my head, there's an expectation of being closer to fully billable pre- achieving the +lead trait

alex-min:
:+1:

roles/engineer.md
lukemorton: Yeah I changed this here to provoke this conversation actually. I'm not sure how to document this difference in expectation.

chrisblackburn: The current way we bill AaaS teams requires a full-time ("good faith") billing commitment. Not to say that's set in stone necessarily, but would require a billing rethink. 

Given where we're currently hiring, do we have an expectation that everyone joining the team could be fully billable in two weeks? I think the difference was when we were focusing our hiring around less experienced engineers?

lukemorton: Yes I think on project work we can probably say most people will be fully
billable. It's well known that Support isn't at such utilisation, not just
down to experience level but the bitty type of work that can sometimes
occur. Of course as we get better at using iterations to ship Support work
then I guess we will see utilisation go up too.

PR #16, opened 2016-08-16
Introduce outline of new traits

lukemorton:
This is based on previous conversations between myself and various colleagues.

The aim here will be to introduce both needs and concrete expectations so that people know exactly what we're looking for and can easily be measured. At the moment I've only added needs to trigger a conversation around these traits.

Currently taking input on:

 - Other traits that we think would be valuable to introduce now
 - Suggestions on expectations for traits
 - Updates to needs for the traits listed here

### DIFF
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -20,6 +20,36 @@ We expect someone in the role:
 
 ## Engineer traits
 
+### +Advocate
+
+We need someone:
+
+ - to represent the company by speaking in public
+ - to represent the company on social media
+ - to represent the companies hiring interests publicly
+
+### +Architecture
+
+We need someone:
+
+ - to lead implementations of application and system level design
+ - to actively promote the evolution of existing application architecture
+
+### +Customer
+
+We need someone:
+
+ - to regularly communicate with customers
+ - to facilitate communication between their team and customers
+ - to encourage improvements in how we communicate with customers
+
+### +Delivery
+
+We need someone:
+
+ - to regularly improve our process for delivering software
+ - to ensure the done done completion of tasks
+
 ### +Lead
 
 We need someone:
@@ -34,3 +64,10 @@ We expect someone in the role:
 - to need less than 1 day per week of Technical Director involvement in their deliveries
 - to be mentoring middleweight developers on their team within 2 months
 - to be improving our practices across the business within 3 months
+
+### +Mentor
+
+We need someone:
+
+ - to provide a mentorship role to a colleague
+ - to encourage learning across the company
--- a/roles/README.md
+++ b/roles/README.md
@@ -71,3 +71,21 @@ to offer up extra activities or even a new role.
 
 We want to foster a culture where everybody feels empowered to do their role and
 as they feel comfortable and ready push in any direction they wish.
+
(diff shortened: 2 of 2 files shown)


roles/engineer.md
```diff
@@ -34,3 +64,9 @@ We expect someone in the role:
 - to need less than 1 day per week of Technical Director involvement in their deliveries
 - to be mentoring middleweight developers on their team within 2 months
 - to be improving our practices across the business within 3 months
+
+### +Mentor
+
+We need someone:
+
+- to be helping colleagues and company friends develop their skills
```
craigjbass: This is very broad

lukemorton:
It might be that +Lead and +Mentorship are too similar and we only go ahead with one for now.

Previous thinking was that +Lead trait has already started appearing amongst teams and is typically a more experienced member of the team who helps push teams forward, provides more of a scrum/flow master role, pulls everyones socks up in terms of code standards, etc.

+Mentor is more about helping those in your team learn new skills or helping people in the wider company or even outside the company.

### +Mentor

We need someone:

 - to provide a mentorship role to a colleague
 - to encourage learning across the company

roles/engineer.md
```diff
@@ -20,6 +20,36 @@ We expect someone in the role:
 
 ## Engineer traits
 
+### +Advocate
+
+We need someone:
+
+ - to represent the company by speaking in public
+ - to represent the company on social media
+ - to represent the companies hiring interests publicly
```
chrisblackburn: company's

craigjbass:
@lukemorton is the outcome to have a broad approach to each role, or a very narrow approach

I would say +Coach would make sense instead of +Lead, but then this would not predicate a tech background. 

If the tech knowledge is split out this could then be +Mentor but then this predicates some mentoring ability in one or more particular stacks. This might be a large barrier to entry if we want to go down the Polyglot approach and have lots of stacks. Potentially each +Mentor could be in a particular stack, but then to have a role around a stack would potentially become a business barrier to entry to just picking up a new stack as there is already inertia around other stacks.

+Architecture is interesting as this should encapsulate transferable knowledge, and be stack agnostic.

I would say I can mentor on generalised knowledge that can apply ATDD to Ruby, PHP, Java, Kotlin but it would be a joke for me to attempt to mentor someone on Rails or Spree at this time.

lukemorton:
I would say all traits will be agnostic to a particular technology.

craigjbass:
Do you expect everyone to get +Mentor if they are capable of mentoring

lukemorton:
I think we will layout expectations as per the Engineer role and existing traits which will be quantifiable to the point we can say yes this person exhibits this trait or not.

If everyone fulfils the criteria of the +Mentor then they are indeed +Mentor.

craigjbass:
Perhaps a narrow approach would be highly beneficial

lukemorton:
Okay I feel we've got a nice mix of traits at the moment. I've left +Lead in for now, however it looks more likely that more specific traits will end up superseding the need for +Lead.

Merging this PR in. Next couple of PRs will handle +Lead and expectations for traits.

PR #17, opened 2016-08-19
Basic instructions on segmenting new leads

chrisblackburn:


lukemorton:
Would probably be good to link to the master contact list. Have been linking to trello and google docs in this repo even though public as all the links are locked down.

PR #18, opened 2016-08-22
1. Replace +Lead with idea of emerging Engineer Leadership

lukemorton:
Previously we used +Lead to encompass the properties that are required to ensure engagements go smoothly. As we developed the traits we wish to encourage we've seen more and more of the expectations of +Lead fall under other traits and therefore feel +Lead isn't a trait in and of itself.

Engineer Leadership should be recognised, and is useful when forming teams. A primary point of contact and someone responsible for the commercial success of an engagement is required within a team in order for that team and engagement to run itself. However rather than grouping expectations around +Lead we recognise Leadership as an emergent property when an Engineers begins to exhibit more and more traits.

### DIFF
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -56,3 +56,13 @@ We need someone:
 
  - to provide a mentorship role to a colleague
  - to encourage learning across the company
+
+## Engineer Leadership
+
+Traits recognise value that Engineers can provide customers and the company. They are hats and as Engineers develop their skills, Engineers will begin to wear more of those hats. Leadership is a quality that emerges out of individual traits. The longer an Engineer wears a trait the more of a leader in that area they will become.
+
+Engineer Leadership is a property that emerges when an individual becomes a leader across multiple traits and starts to perform more vital roles within engagements with customers.
+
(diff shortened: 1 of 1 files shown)


roles/engineer.md
```diff
@@ -50,24 +50,19 @@ We need someone:
...
-- to be improving our practices across the business within 3 months
+Traits recognise value that Engineers can provide customers and the company. They are hats and as Engineers develop their skills, Engineers will begin to wear more of those hats. Leadership is an emergent quality that develops under individual traits. The longer an Engineer wears a trait the more of a leader in that area they will become.
 
-### +Mentor
+Engineer Leadership is a property that emerges when an individual becomes a leader across multiple traits and starts to perform more vital roles within engagements with customers.
 
-We need someone:
+For example, when an Engineer becomes a leader in +Customer and +Delivery they will naturally begin to be the primary point of contact to the customer for the engagement, not just in regards to communication and agile process but also in other commercial areas. Relationships will naturally evolve between Engineers fulfilling more skills in the team and the customer, this seniority will be recognised. As these relationships develop with customers, more commercial level responsibility will fall on this emerging leader.
```
chrisblackburn: Is 'fulfilling skills' right?

chrisblackburn: 'will fall' or 'should be assumed'? responsibility being taken rather than given, and all that

lukemorton:
@chrisblackburn Improved the wording here.

PR #32, opened 2016-09-30
Rearrange headings, with intention to make them feel more evolvable

craigjbass:


lukemorton:
I like the idea of Guides. We could probably roll up policies into Guides.

Not too sure on the wording "Team norms". Would prefer something feel like "How we do things", "The Made Way", "Engineer Documentation", or "Engineer Wiki". Team norms does do a better job of describing team ownership over the process though.

Interested to get as many peoples thoughts as possible on this.

craigjbass:
I like Team norms as it encapsulates the idea that they are evolutionary, and it's somewhat well used in the agile community to mean that. Not sure we should create our terminology unless we can do a better job of it

craigjbass:
@lukemorton I've fixed up the directory structure as well

PR #36, opened 2016-10-05
Proposal for decision making (WIP)

craigjbass:


lukemorton:
I actually agree with all of this :)

Only thing in my mind is making sure we have the ability to escalate from problem from dealing one on one if things got really bad. Of course this is a rarity but I think it's important to have an escalation procedure for disagreements if they're affecting others negatively.

craigjbass:
@lukemorton yeah, this is the part which I think needs to be discussed and adopted first as a team norm. 

In theory, conflict resolution should initially be attempted to be resolved 1-2-1 between the parties who are at conflict, this prevents noise, it should only be escalated when that fails, or there is an extreme lack of safety. 

Decisions made without following the advice process are "illegal".

craigjbass:
Feel like this is resolved by #54

PR #38, opened 2016-10-23
Added C# marking guide, fixed wording on values page

paulmorrishill:
I've avoided specifying any particular .net libraries to use as there are quite a few different ones that can be argued for and it would be wrong to exclude a candidate due to their particular personal preference. 

I think the wording reads better on the values page this way.


lukemorton:
Perhaps we should rename the C# marking guide to `language_agnostic_marking_guide.md` seeing as there aren't any direct references to the language?

guides/hiring/remote_test/csharp_marking_guide.md
```diff
@@ -0,0 +1,28 @@
+# Marking guide
+
+60 marks in total. There are no half marks. 80% (48 marks) required to pass.
```
lukemorton: What marks did we remove as they were too specific? Perhaps there are some language agnostic alternatives?

lukemorton:
Looking good, thanks for this!

guides/hiring/remote_test/csharp_marking_guide.md
```diff
@@ -0,0 +1,28 @@
...
+ - [2] The timeline appeared below the form
+ - [3] The timeline is ordered in descending chronological order
+ - [2] Each item in the timeline contains a message
+ - [2] Each item in the timeline contains the time of submission
+ - [2] Each item in the timeline contains a way to delete the post
+ - [3] Message can contain newlines, new lines are rendered
+ - [3] Time is displayed in format of "20 minutes ago", "5 days ago"
+ - [3] Clicking delete button results in item being deleted
+ - [3] Clicking delete button results in message being displayed
+ - [3] Git was used to version control
```
craigjbass: If we make this the language agnostic marking guide, perhaps we should drop specifying "Git"... I personally wouldn't mind if someone pushed up a Mercurial repo, for example

guides/hiring/remote_test/csharp_marking_guide.md
paulmorrishill: > [2] Is the app built with Rails?
> [3] Rubocop is installed and is passing

Not all languages have tools of this nature although to be fair some languages absolutely need them - eslint for JS for example.

> [2] Scaffolding is not used for models
> [2] Scaffolding is not used for views
> [2] Scaffolding is not used for controllers

> [2] RSpec is used

You could specify a requirement for a BDD test framework but I'm not sure that really matters as long as the tests are readable and easy to understand. 

> [2] Capybara is used

Not sure about how absolutely necessary a gherkin syntax is for a project of this size or in general.

> [2] shoulda-matchers is used

I also added a 1 pointer about a nicely readable assertion library being used.

The thing I would be weary of is that most of these are more around getting the tooling set up, which is usually a one off task at the start of the project. I think the more important thing is making sure that the code is up to scratch quality wise and architected competently as that is the bulk of what they'll be doing day to day.

guides/hiring/remote_test/csharp_marking_guide.md
lukemorton: Yeah could change it to "Code is version controlled"

guides/hiring/remote_test/csharp_marking_guide.md
paulmorrishill: > We pretty much recommend static code analysis for....

Fair.

> To see if a candidate can Install and set up a command line tool

God I hope so.

> This isn't gherkin syntax. Capybara provides browser drivers

My mistake, I see this is the ruby interface for selenium I quickly skimmed the capybara readme and misread. Pretty much all languages have a library for selenium so agreed.

lukemorton:
Very happy with this 👍

lukemorton:
Would like to await one or two more reviews. Particularly interested in opinions regarding the difference between the rails marking guide and the language agnostic one.

guides/hiring/remote_test/language_agnostic_marking_guide.md
```diff
@@ -0,0 +1,32 @@
...
+ - [2] Feature test: successful submission are passing
+ - [2] Feature test: unsuccessful submission are passing
+ - [2] Feature test: message display are passing
+ - [2] Feature test: message deletion are passing
+ - [3] Unit test: both validation rules are tested and passing
+ - [2] A nicely readable assertion library is used
+ - [3] README contains instructions for running application
+ - [3] README contains instructions for running tests
+ - [3] A linter is set up and passing for all languages used in the project
+ - [3] There are browser based tests using selenium or similar, all tests pass
```
alex-min: personnally not a fan of full browser-based tests myself as a concept but it's a matter of preference

lukemorton: Yeah we typically use capybara with Rack::Test which avoids the slowness of selenium type tests. I think as a language agnostic concept, saying selenium will help people understand what we are looking for.

craigjbass: I think it's quite good to show that a dev knows how to use selenium... You still need a sprinkling of E2E smoke tests on top, generally around the areas of a system that add the most value. Where it becomes a problem is when you use slow E2E tests to TDD out actual logic. IMO that's not what Selenium is for.

PR #50, opened 2017-01-09
Evolution of our recruitment process

RoryMacDonald:


lukemorton:
Looking good so far. Might be nice to condense whole process into a single README.md though so it's easy to consume. If there are more resources for particular techniques they could be linked off to.

PR #53, opened 2017-01-18
Add trait achievement heading

RyanMacG:
One of the most difficult things about traits at the moment is knowing when their done and how to decided that.

This is a chance to boot up a visible discussion for ideas around this.

### DIFF
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -94,6 +94,8 @@ We expect someone exhibiting the trait:
       - helping to onboard new hires to the traits system
       - checking in with colleagues regarding their progress
 
+### Achieving a trait
+
 ## Engineer Leadership
 
 Traits recognise value that Engineers can provide customers and the company. They are hats and as Engineers develop their skills, Engineers will begin to wear more of those hats. Leadership is a quality that emerges out of individual traits. The longer an Engineer wears a trait the more of a leader in that area they will become.


RyanMacG:
We've started to address this by increasing visibility (via Trello) of what people are doing and have done with regards to their traits.

craigjbass:
Feel like #54 + http://www.holacracy.org/badge-based-compensation-app/ are a sensible path to take

RyanMacG:
@craigjbass in that situation who makes the decision for the first person though?

craigjbass:
The badge app defines a policy that you must adopt during a holacratic governance meeting to create a Badge Bootstapper role. They do their best to measure the first achiever of a badge/trait against criteria..

craigjbass:
A snippet...


### Add Role: Badge Bootstrapper

#### Purpose:

Highly discerning and conservative Badge grants to bootstrap just enough people into new Badges
Accountability:

Assessing Partners for Badges upon their self-nomination, when there aren’t yet enough people who actually hold the Badge and work with the Partner who can reasonably assess

craigjbass:
### Add Policy: Earning & Losing Badges

**If no mechanism is specified along with a Badge definition for how a Partner gains that Badge, then, as a default mechanism, the following will apply:**

A Partner will be granted the Badge when at least two people who already hold the Badge assert that they have clearly and repeatedly seen the candidate demonstrate the skill or capacity described by the Badge. However, if there aren’t enough people who actually hold the Badge and have worked with the candidate to assess that, then the @Badge_Bootstrapper’s may instead grant the badge using any evidence they see fit, as long as a majority of the @Badge_Bootstrapper’s advocate for the grant and none actively advocate against it. If the candidate for this grant is also a @Badge_Bootstrapper, he/she shall not count for the purpose of assessing his/heir own Badge nomination.

**If no mechanism is specified along with a Badge definition for how a Partner loses that Badge, then, as a default mechanism, the following will apply:**

The Badge grant will automatically expire two years after grant, unless renewed via the same mechanism required for the initial grant. Further, the Badge grant may be revoked sooner by any two people who already hold the Badge asking for a re-qualification, and the Partner failing to promptly re-qualify for the Badge using whatever mechanism is required for the initial grant of the Badge.

craigjbass:
Think it alludes to the fact that there are multiple gatekeepers for badges.

craigjbass:
I'm wondering whether Clean Architecture should be a separate trait, with Rails, React and any other paradigm we see as distinct enough having a separate trait to recognise them

chrisblackburn:
^ maybe useful to spin this thing about separate traits in to another convo (I'm interested in commenting, but wonder if it's OT to the original question)

RyanMacG:
I don't currently see the advantage to it and it's not really the focus of this PR

chrisblackburn:
Some thoughts on the original:

- A regular forum (ideally not a meeting!) for reviewing progress towards a trait - i.e. something that happens to a schedule so it doesn't get forgotten
- Maybe when people feel they've achieved their trait goals, they put a case forward to that effect, perhaps citing support from others
- Something in the CF sheet that provides a structure for tracking progress
- My current view is the Holacracy stuff is interesting, but question it snowballing this current aim (i.e. it opens more questions than it answers at this point)

RyanMacG:
@chrisblackburn when we tried to apply points 1+2 there was a suggestion that there should be a set amount of time to be evidencing the trait being performed further

RyanMacG:
I don't agree with the timescale idea but wanted to highlight it

lukemorton:
Yeah maybe it just requires evidence being put before a committee? (Committee is my favourite unpopular word of the month.)

We'd still need some agreed upon requires to be interpreted by said committee.

lukemorton:
How can we move this forward?

craigjbass:
Does anyone see a reason why Luke's proposal would cause harm?

chrisblackburn:
Only the negativity I associate with the word committee or the notion of "design by committee" - https://en.wikipedia.org/wiki/Design_by_committee :-)

I think it's important that attaining a trait is seen as a Big Deal, and so the process through which a trait is awarded, while definitely doesn't need to be heavyweight or long-winded, does need an amount of rigour about it. For example, some people may be less comfortable with saying "actually, no, I believe you need more work in this area to really attain this trait". Could that be a challenge?

craigjbass:
@chrisblackburn do you have a counterproposal?

chrisblackburn:
I think I have something that builds on that proposal, really:

- Absence of the word committee :-)
- A light pre-defined template for showing your belief that you've attained a trait (some bullet points listing achievements or something?)
- Present (not ala powerpoint) that reasoning to 3 people, all of whom need to agree that this trait has indeed been achieved against some pre-defined criteria
- One of those people should be a Director, which perhaps serve the purpose that at least 1 of the 3 feels safe enough to not agree
- Big celebration / constructive feedback on what is missing to attain the trait

Slightly separately, I wonder if some interim checkpoints may be beneficial - some way to get confidence that you're progressing well against your trait

craigjbass:
Feel like measuring success against (measurable) objective principles are the only way to fairly award a Trait.

* Where there are no pre-existing individuals who hold a specific Trait, the individual who is being assessed for a Trait award must present a set of criteria to 3 people. 
  * The goal is to obtain agreement that those criteria are valid criteria for the Trait to be met.  
* The individual under assessment must then present Evidence that they have met those criteria to those 3 people.
* The 3 people must be satisfied that the Evidence meets the outlined criteria for the trait to be awarded. 
  * There is another opportunity for the criteria to be modified, but only if all parties are satisfied it makes sense to do so.

**Problem:** what if two people are working towards a trait, and disagree on the criteria. Does the criteria agreed upon affect all individuals working towards a trait?

chrisblackburn:
Could criteria be proposed and agreed upon via PR? Needs to be limited to those who are working toward that trait, or open to all?

The traits were originally introduced as a way of expressing skills and behaviours that support the company's mission. Feels like others may have interest in shaping acceptance criteria?

RyanMacG:
I'm concerned that if we don't define this to at least some degree we'll be setting ourselves up for situations where people feel the goalposts are being moved based on their pre-conceptions.

craigjbass:
@RyanMacG Agree, it's not good on motivation to have continuously moving goal posts.

lukemorton:
Okay a few questions to move this forward:

- Is the listed criteria, at least for +Mentor, fit for purpose of identifying whether we think someone is exhibiting a trait or not
- Once we have a list of criteria, is there a minimum amount of time it should be met for before awarding a trait? You may want this so we don't award after 2 weeks of exhibiting a trait and then need to take away a trait after effort is dropped
- Do we want to avoid traits being taken away at all?
- Is everyone happy once we have some quantifiable criteria, we just have 3 people review whether the trait would be awarded, with one director amongst them?

davidwinter:
I think we need to move this forward. After almost two weeks, comments have dried up on this and it doesn't feel like we've got a decent consensus. Maybe this isn't the best forum to discuss and decide on this. Could we use the comrado this week to get this wrapped?

If not, is there any issue with us trialing the above suggestions? For at least +mentor? I like the idea of a panel of three to discuss and decide if it's been achieved, and I think the criteria is best defined for +mentor compared to the others currently. We shouldn't be afraid of modifying these in the future if we think they need to, or the panel idea of awarding the traits, if that doesn't work out.

Some people are putting significant effort into working towards their traits, so feel it's unfair for us not to come to a decision on this. Otherwise I can see it feeling as though you're following a carrot on a stick.

Once we've made this decision, I think there's plenty of other things to discuss. My thoughts on some of the other points:

Traits being taken away... I can see why this might be needed. But also, would it be good that once you've achieved the trait, it's more of a demonstration of what you're capable of, and that trait sticks with you, kind of like a scout badge? Maybe this needs to be handled separately as it's been mentioned traits will be used in pay reviews.

Should people have a single active trait? But be able to swap out the active trait for another? Once you've achieved a trait, and switched to another, it becomes an 'Inactive' trait - but one that you've still achieved? But again, I imagine this opens a whole other can of worms with how traits are linked to salary.

craigjbass:
I see activities performed during the period of time working towards a trait as mostly actions to prove that the understanding, knowledge and practical skill is present. After achieving the trait, the individual could choose to focus his/her efforts elsewhere whilst retaining the trait that they achieved. 

I see the role of +Architecture, for example, overlapping massively with +Mentor. I actually see the mentoring role of someone helping other's with Architecture much more difficult as you're mentoring whole teams, instead of individuals on a 1:1 basis. 

For example, if a team need to be made aware of a particular architectural principle, pattern or the evolutionary strategy a +Architecture will need to:
- Have the experience/knowledge to know when a pattern, principle or evolutionary strategy is contextually appropriate
- Have the ability to match-make another's knowledge with a problem, i.e. Bob has knowledge in x. Alice has problem y. +Architect can see that Bob's knowledge of x can solve Alice's problem y  
- Also have good leadership skills, empathic arguments, good salesmanship and negotiation skills

A +Architecture becomes stronger every time he/she has:
- Added to his/her knowledge of patterns/principles/strategies/tools
- Improved his/her mentoring & leadership skills
- Added to his/her experience around cost/risk/reward understanding of previously mentioned 

I see a +Architecture as a catalyst for ensuring customers get the most amount of primary value (the software is cheap to adapt to changing/new/future requirements) out of their systems, whilst ensuring strategic decisions to increase primary value do not hold back delivery of secondary value (the software meets current requirements). 

*My point being...* this is not something you stop knowing overnight, it's also not something you simple stop giving outwardly to a team. I think +Mentor, for example, will begin to make more sense once we have a very clear experience gap within the team, as individuals will need continuous Mentoring while even being billable. In the same way we share domain knowledge of how customer's software deliveries are put together.

My Opinion: Think the hardest part of Traits is it asks are a very difficult inward question from all of us, is y better at z than me? And does y provide more value in z field than me consistently on deliveries? 

Think it's important to remember that it's not about being better at deliveries, it's about being recognised as serving a certain business function because significant time has been put in on the awardee's part to ensure they have sufficient skill & deliver real value to the business in that area.

PR #61, opened 2017-03-08
Add link to first boot

SebAshton:


naliwajek:
It's awesome and it works 👍

PR #73, opened 2017-05-24
Clarify the overlap between Untracked Holiday and Flexible working

craigjbass:


benefits/untracked_holiday.md
```diff
@@ -4,6 +4,10 @@ We're building a high-performance organisation that delivers great software prod
 
 We have adopted an untracked holiday policy, where each team member is given the flexibility to take whatever holiday they need. We believe this policy will help us to perform at the highest possible level. 
 
+It is important that you and your colleagues feel as though that holiday is highly available. That holiday is an easily accessible commodity that enables everyone to get the necessary rest that they need. We intend untracked holiday to prevent burn-out in situations, where individuals feel that they need to "save up" holiday days, or plan elaborate strategies around using your holiday allowance "effectively" later in the year, rather than taking holiday when needed most.
+
+Untracked Holiday is not meant as a way to regularly take large portions of time off work e.g. 1 week every month, or 4 months off straight. Please see our [Flexible Working](flexible_working.md) as the alternative to Untracked Holiday in those situations.
```
RyanMacG: Is the second part of this true? What is the upper limit of the untracked holiday?

craigjbass: Since we generally work in pairs, if I was to observe someone being regularly unavailable within my delivery team. I would feel frustrated and disrespected by them. As it would impact my need to deliver value to our customers and my need to ensure my pair could take over from me if I needed to take holiday.

I'm not sure what the number is exactly, but those needs should be met IMHO

RyanMacG: Of course, I'm just not sure if we have an upper limit really? Longer holidays obviously get floated to the wider company to cover shifting between teams/engagements but I've never heard anyone discuss a set point where they wouldn't accept a holiday.

benefits/untracked_holiday.md
```diff
@@ -4,6 +4,10 @@ We're building a high-performance organisation that delivers great software prod
 
 We have adopted an untracked holiday policy, where each team member is given the flexibility to take whatever holiday they need. We believe this policy will help us to perform at the highest possible level. 
 
+It is important that you and your colleagues feel as though that holiday is highly available. That holiday is an easily accessible commodity that enables everyone to get the necessary rest that they need. We intend untracked holiday to prevent burn-out in situations, where individuals feel that they need to "save up" holiday days, or plan elaborate strategies around using your holiday allowance "effectively" later in the year, rather than taking holiday when needed most.
```
chrisblackburn: feel that holiday
OR 
feel as though holiday

(probably the former for my taste)

benefits/untracked_holiday.md
chrisblackburn: We've never really been close to a situation where the policy has been abused. But as noted, if you wanted to take, for example, 4 months off, it could be that an unpaid sabbatical is more appropriate. Whether this needs explicitly calling out, I'm not sure. Again, never been close to a situation where it felt the "spirit" of the policy was being pushed.

RyanMacG: We don't have a policy around sabbaticals though, do we? I'm wondering if it's better to not have the example unless we want to clarify around that area?

craigjbass: I feel anxious when I consider having no example as it doesn't meet my need for unambiguousness of what is expected of me and of others.

RyanMacG: I can understand that, maybe this is something we need to work out in a separate PR

PR #74, opened 2017-05-26
Inclusive team lunches

craigjbass:
While this is specifically about Veganism, I feel it is a much broader topic of diversity that we need to be aware of.


lukemorton:
@craigjbass let's merge?

PR #81, opened 2017-08-03
Maternity, paternity and shared leave

RoryMacDonald:
Some thoughts on maternity / paternity / shared parental leave

### DIFF
--- a/guides/welfare/parental_leave.md
+++ b/guides/welfare/parental_leave.md
@@ -12,6 +12,6 @@ We provide two weeks Paternity Leave at full pay and the option of shared parent
 We provide two weeks Shared Parental Leave at full pay and up to 37 weeks at statutory pay. The rules around SPL are fairly complex, government guidelines are available here: [https://www.gov.uk/shared-parental-leave-and-pay](https://www.gov.uk/shared-parental-leave-and-pay).
 
 ### Other Considerations
-We offer a number of additional company benefits which may be attractive to new parents, including [Remote Working](benefits/remote_working.md), [Untracked Holiday Allowance](benefits/untracked_holiday.md), [Flexible Work Weeks](benefits/flexible_working.md) and [Flexible Working Hours](benefits/working_hours.md). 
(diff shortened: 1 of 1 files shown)


idabmat:
This is much clearer :+1:

lukemorton:
Although I think this is what we currently offer I don't think it's a fair package. Seeing as we already allow staff to take 4 weeks holiday we could at least offer 4 weeks full pay if not 6.

guides/welfare/parental_leave.md
```diff
@@ -1,7 +1,17 @@
-# Parental Leave
+# Maternity, Paternity & Shared Parental Leave
 
-Our current parental leave package is two weeks at full pay followed by statutory allowance.
+When you welcome a new child (birthed or adopted), Made Tech encourages you to take leave, to welcome the child into the world.
```
sowasred2012: Lots of "welcoming" - maybe:
When you welcome a new child (birthed or adopted), Made Tech encourages you to take leave and spend time introducing your child to the world.
?

sowasred2012:
Not sure how much of this is down to Government mandate or a per company basis - but would love to see the UK adopt [the Swedish approach](https://www.theguardian.com/world/2015/may/28/swedish-fathers-paid-paternity-parental-leave).

guides/welfare/parental_leave.md
RyanMacG: I'd recommend `introducing your child to the family` or something, adopted children generally older so have already see the world a bit. Just a thought

RyanMacG: For context:

> 5% (3,540) of children looked after on 31st March 2016 were under 1 year old
13% (9,140) were aged between 1 and 4 years old
20% (14,090) were aged between 5 and 9 years old
39% (27,220) were aged between 10 and 15 years old
23% (16,460) were aged 16 and over

From - https://www.adoptionuk.org/press-media/adoption-facts-and-figures-england

chrisblackburn:
@lukemorton r.e. the holiday pay, I think you still accrue holiday allowance while on leave (https://www.nidirect.gov.uk/articles/entitlements-during-statutory-maternity-leave)

lukemorton:
@chrisblackburn Yeah that would be the additional cost to the business compared to using holiday instead of mat/pat leave.

PR #82, opened 2017-08-08
Academy Role

chrisblackburn:


roles/engineer.md
```diff
@@ -18,6 +16,21 @@ We expect someone in the role:
 - to be autonomously delivering features of a small nature within 3 months
 - to be autonomously delivering features of a medium nature within 6 months
 
+
+## Academy role
```
craigjbass: Not sure about the name of this role `Academy`. `Apprentice Engineer` ?

lukemorton: `Academy Engineer`

craigjbass: What does the qualifier `Academy` mean? Would it require further explanation?

roles/engineer.md
```diff
@@ -18,6 +16,21 @@ We expect someone in the role:
...
+
+## Academy Engineer role
+
+We need someone:
+
+- to train to be ready to join a delivery team following graduation
+
+We expect someone in the role:
+
+- to be autonomously delivering features of a micro nature before graduation
```
RyanMacG: Have we got any better language for this? Agree with what you said earlier about it being unclear.

craigjbass: "low complexity" => implications on size of domain, customer's understanding of their own domain / needs / desires? Solution for the problem is of low technical complexity - requires very few moving parts i.e. data is already in near-enough correct format, need presentational work?

roles/engineer.md
```diff
@@ -15,8 +13,23 @@ We expect someone in the role:
 - to be fully billable after 2 weeks on-boarding
 - to be involved in practice improvement discussions within 3 months
 - to be actively practicing modern software craftsmanship practices such as TDD within 3 months
-- to be autonomously delivering features of a small nature within 3 months
-- to be autonomously delivering features of a medium nature within 6 months
+- to be autonomously delivering features of modereate complexity within 3 months
```
craigjbass: `moderate`

Issue #76, opened 2017-05-30
Paternal leave


camillavk:
Currently the paternal leave is a little unclear in the handbook. It says:

"Our current parental leave package is two weeks at full pay followed by statutory allowance." 

A couple of questions I have about this is;
1) Does "paternal leave" cover both maternity and paternity leave as the same thing?

2) What exactly does "followed by statutory allowance" entail?
On the gov.uk [paternity](https://www.gov.uk/paternity-pay-leave/leave) site paternity leave is stated to be: "You could get either 1 or 2 weeks". So does our paternal leave package mean 2 weeks full pay + 1 week statutory, or 2 weeks full pay + 2 weeks statutory ?
On the gov.uk [maternity](https://www.gov.uk/maternity-pay-leave/leave) site, the statutory allowance is a pretty simple 52 weeks, but does out paternal leave package mean we get 2 weeks at full pay on top of that? (i.e. 54 weeks) ?

I think maybe separating this out in maternity and paternity leave so that it is clearer might be useful?

lukemorton:
Personally I'd hope to have parity between paternity and maternity leave but we should aim to have this explicitly made clear.

camillavk:
I think having parity between paternity and maternity leave is a good idea, but I think this needs to be clearer, as I think it's a little unclear at the moment exactly how much leave you would actually get in either situation

lukemorton:
Yes, definitely agree. So far we have provided paternity leave to 2 members of the team. Let me clarify the details around this to let you know the route we've taken so far.

RoryMacDonald:
Yea, there are guidelines on shared parental leave / pay on GOV UK too:

https://www.gov.uk/shared-parental-leave-and-pay

camillavk:
Ah ok, I think the link you posted @RoryMacDonald makes a little bit more sense...I think maybe at the very least we can add the link to the handbook? For someone who doesn't know much about how paternal leave works in the UK that goes a little way to explaining it more

lukemorton:
So I think we need to update the handbook into a decent state, to reflect the various options around statutory. I'd also like to look at how we can clearly provide parity (as much as possible) for paternity and maternity.

I'll certainly be happy to investigate this when I'm back from holidays.

RoryMacDonald:
I've created a small PR to try to clarify things: https://github.com/madetech/handbook/pull/81/files

Good to get some feedback on this.

lukemorton:
[...]

PR #85, opened 2017-11-02
Breakdown pay for maternity, paternity and shared parental leave

lukemorton:
It makes it a little easier to understand what we offer by including break down of pay.

SPL is complex and therefore employees should consult gov.uk pages we already link to. Our employee wellbeing helpline can also provide more information around this.


lukemorton:
Relates to #76

PR #86, opened 2017-11-17
Clarify inclusion around Friday Lunches

lukemorton:
This change should make it clear that generally we want everyone to eat together, and lunch must always be inclusive to dietary requirements.

In order to have a wider variety of restaurants on our rota, including vegan and meat only restaurants, we are clear that on those weeks if a dietary requirements are excluded then those excluded may go to a second restaurant of their choice and expense it too.


davidwinter:
Mention something about critical project work/meetings that run over the normal lunchtime, and that these can be expensed?

lukemorton:
I think they should get more than just a free lunch :)

Issue #80, opened 2017-07-14
Escalation framework


lukemorton:
I'd like to write a framework/document for the following:

- What to do when I need to escalate an issue? I.e. disagreement between staff, disagreement with leadership, when someone has been wronged, etc.
- Who can I talk to when I have a problem?
- Is there an external body I can receive help from?

craigjbass:
How would we approach this from a Self-Management perspective:

* It’s impossible to change other people. We can only change ourselves.
* We take ownership for our thoughts, beliefs, words, and actions.
* We don’t spread rumors.
* We don’t talk behind someone’s back.
* We resolve disagreements one-on-one and don’t drag other people into the problem.
* We don’t blame problems on others. When we feel like blaming, we take it as an invitation to reflect on how we might be part of the problem (and the solution).
* We focus on strengths more than weaknesses, on opportunities more than problems.

# Steps

1. First, methods are developed to help people bring tensions to the surface.
2. Second, clearly defined conflict resolution processes are available to help people safely confront each other when needed.
3. Third, most self-managing organizations train every new recruit in conflict resolution and interpersonal skills.

# For example:
In the first phase, the two people sit together and try to sort it out privately.
If they can’t find a solution agreeable to both, they nominate a colleague they both trust to act as a mediator. The mediator doesn’t impose a decision. Rather he or she supports the participants in coming to their own solution.
If mediation fails, a panel of topic-relevant colleagues is convened. Again the panel does not impose a solution.
If resolution is not found, the founder or president might be called into the panel to add to the panel’s moral weight (but again, not to impose a solution).

# Training in Interpersonal skills
Because effective conflict resolution skills are so central in self-managing organizations, many organizations train all their colleagues in interpersonal skills to enable them to deal gracefully with conflict. Generally in their first weeks at work, new hires are given foundational training including: self-management, deep listening, dealing constructively with conflict and creating a safe environment. For instance, companies like ESBZ and Buurtzorg train colleagues in Nonviolent Communication developed by Marshal Rosenberg.

[source](http://www.reinventingorganizationswiki.com/Conflict_resolution)

PR #88, opened 2017-11-18
Add information on "Raising an Issue"

lukemorton:
Related to addressing #80 and following on from the discussion during comradrospective this week, I've put together a first draft of information to empower the Made Tech team to be able to resolve issues both informally and formally.

All staff should take the following steps, if possible, to resolve their issue:

- Have a conversation with the person involved
- Raise feedback via Continuous Feedback
- Ask a colleague to intervene on your behalf
- Seek Third Party Advice
- Mediation
- Formal Grievance Procedure
- Arbitration Tribunals and Civil Courts

Not the loveliest subject but I really think it's important we outline procedure so everyone feels empowered and protected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ We're a company that moves fast, so we're going to need everyone to help us keep
 * [Sick Leave](guides/welfare/sick_leave.md)
 * [State of Mind](guides/welfare/state_of_mind.md)
 * [Employee Assistance](guides/welfare/employee_assistance.md)
+* [Raising an Issue](guides/welfare/raising_an_issue.md)
 * [Parental Leave](guides/welfare/parental_leave.md)
 
 ### Guides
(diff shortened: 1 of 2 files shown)


craigjbass:
I think there are two parts to this, escalation for things that are non-severe and highly severe things. We need to support people in both categories.

I feel the steps listed escalate quickly. I think there needs to be more internal steps first.

For example, while nobody is a trained mediator, I’d expect the second step to be to encourage someone to facilitate a meeting between yourself and a person you are at conflict with.

lukemorton:
I can see that. I think it's really important for most issues to be solved in the first few steps. It does escalate but I'd expect 90% of problems to be solved by the first step, 9% solved by the second, .9% solved by the next, etc.

I feel like "Ask a colleague to intervene on your behalf" is the step you talk about. Do you think we could add another paragraph to that section (or others) related to encouraging facilitation?

craigjbass:
I think there is a difference between intervening and facilitating.

lukemorton:
You think the facilitation could be added to "Have a conversation with the person involved" as an alternative then?

PR #94, opened 2017-12-20
[WIP] Add Senior Engineer role expectations

lukemorton:


roles/senior_engineer.md
```diff
@@ -1,3 +1,49 @@
...
+
+We expect our Engineers to be fully billed on customers deliveries as we are a business trying to make a profit by delivery valuable working software to our customers.
+
+## To be involved in practice improvement discussions
+
+We expect everyone in our team to reflect on our practices and be a proactive voice in suggesting change. We expect you to be reflecting on your own development practices, reflecting on your team and the wider business.
+
+## To be actively practicing TDD
+
+We expect our engineering team to write tests first by default. We expect every feature to be tested. Every Pull Request (or equivalent chunk of work put up for peer review) must contain a test. If for some reason there are extenuating circumstances for not writing tests then more than one engineer should peer review and verify this.
```
craigjbass: I know you said you're not looking for comments, but want to nitpick. Are we expecting `test first` or `TDD`? It's important be clear about the difference. 

TDD is the discipline that we've been focussing on in Learn Tech and test first is not a discipline.

chrisblackburn:
Are there specific things we want to hold Senior Engineers to account on? e.g.

A Senior Engineer is expected to be responsible for the quality of the code delivered by the team
A Senior Engineer is expected to provide valuable and meaningful feedback for Engineers on their teams
A Senior Engineer is expected to lead the technical architecture of complex projects

chrisblackburn:
And (not exactly sure of the measure but..) something about "expert" proficiency in one or more programming languages? In my head at least, the Senior moniker implies some deeper than average knowledge

craigjbass:
@chrisblackburn I'd have said `paradigms`. Think a broad understanding of the `process-oriented` way, `functional programming` way, `object-oriented` way, `imperative` way.

If a Senior Engineer was tasked with introducing dependency inversion in a new language but is in one of these paradigms they should have enough grounding to understand what to look for and how it generally is applied.

chrisblackburn:
Yeah, that makes more sense actually

PR #95, opened 2017-12-20
Add Lead Engineer role

chrisblackburn:


roles/lead_engineer.md
```diff
@@ -0,0 +1,91 @@
...
+
+A Lead Engineer is expected to naturally position themselves as the primary day to day point of contact for the customer. A Lead Engineer is expected to gain the trust of the customer such that they’re seen as the first port of call for all customer contacts.
+
+
+### Ensure customer expectations are exceeded given the budget available
+
+A Lead Engineer is expected to understand the financial constraints, as well as the customer objectives for the delivery. A Lead Engineer is accountable for ensuring the customer’s expectations are exceeded for the available budget.
+
+
+### To be ultimately accountable for the successful delivery of customer engagements
```
craigjbass: Are we saying this is the Delivery Manager is accountable for this? Not sure this is the right split

chrisblackburn: This is on engagements *without* a DM. What do you think the right split is?

craigjbass: Well that's my point, I'd have said that `delivery` itself can only be actioned by the people doing the work i.e. cutting the code.

 A delivery manager - someone who is accountable/responsible for lining up the surroundings of the work cannot be ultimately responsible/accountable for the cutting of code.

Maybe Lead Engineer is always ultimately accountable for successful delivery

chrisblackburn: Is it reasonable to see that cutting code is an important part of the delivery, but far from the only thing that matters?

Maybe it needs clarification that Leads are always ultimately responsible for the successful technical implementation, and on deliveries without a DM are further responsible for the overall success of the delivery?

lukemorton: I'd have the same expectations no matter what team. Otherwise Lead means different things in different contexts.

Why can't we have Lead Engineer being ultimately responsible even with DM?

chrisblackburn: I'm not actually averse to that - though we'll need to redefine the DM role to be clear on this split (the current DM role is clear on accountability for this)

PR #100, opened 2018-01-02
Ask a director if going against holiday guidelines

lukemorton:
Just want clarify that a director should be asked if you want to book holiday outside of guidelines.

### DIFF
--- a/benefits/untracked_holiday.md
+++ b/benefits/untracked_holiday.md
@@ -8,6 +8,16 @@ It is important that you and your colleagues feel that holiday is highly availab
 
 Untracked Holiday is not meant as a way to regularly take large portions of time off work e.g. 1 week every month, or 4 months off straight. Please see our [Flexible Working](flexible_working.md) as the alternative to Untracked Holiday in those situations.
 
+## How long are you taking?
+
+We trust you'll use your best judgement when taking holidays but in case you want some guidance on the kind of notice we expect when you take your holiday:
+
+* **Less than 1 week**<br>We recommend giving at least 2 weeks notice for a 1 day off and closer to 4 weeks notice for more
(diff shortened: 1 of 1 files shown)


benefits/untracked_holiday.md
```diff
@@ -8,6 +8,16 @@ It is important that you and your colleagues feel that holiday is highly availab
 
 Untracked Holiday is not meant as a way to regularly take large portions of time off work e.g. 1 week every month, or 4 months off straight. Please see our [Flexible Working](flexible_working.md) as the alternative to Untracked Holiday in those situations.
 
+## How long are you taking?
+
+We trust you'll use your best judgement when taking holidays but in case you want some guidance on the kind of notice we expect when you take your holiday:
+
+* **Less than 1 week**<br>We recommend giving at least 2 weeks notice for a 1 day off and closer to 4 weeks notice for more
+* **Between 1 and 2 weeks**<br>We recommend giving at least 4 weeks notice for this kind of holiday, ideally more.
+* **Greater than 3 weeks**<br>Please try to provide at least three months notice, so we can prepare for your absence.
```
RyanMacG: Would these all be a bit more readable as

```* **Less than 1 week**  
We recommend giving at least 2 weeks notice for a 1 day off and closer to 4 weeks notice for more
* **Between 1 and 2 weeks**  
We recommend giving at least 4 weeks notice for this kind of holiday, ideally more.
* **Greater than 3 weeks**  
Please try to provide at least three months notice, so we can prepare for your absence.```

rather than using `<br>` inline?

zuren: Does that work with bullet points?

zuren:
Nice change 👍

benefits/untracked_holiday.md
lukemorton: Previously it didn't but I have recently got this to work.

PR #101, opened 2018-01-02
Split friday lunch and drinks

lukemorton:
Splitting these concepts apart as lunch is an inclusive affair whereas drinks are implicitly exclusive to those who like to frequent pubs and such.


benefits/friday_drinks.md
```diff
@@ -0,0 +1,15 @@
+# Friday Drinks
```
adrianclay: Should something be linking to this new file?

lukemorton: Great catch kind sir!

lukemorton:
@adrianclay Done.

PR #110, opened 2018-01-03
Delivery Manager role

chrisblackburn:


roles/delivery_manager.md
```diff
@@ -0,0 +1,63 @@
...
+
+We expect our Delivery Leads to proactively provide regular reporting to the Directors on the health of their delivery.
+
+### Ensure that at all times the team is delivering according to the Made Tech Delivery Standards
+
+We expect our Delivery Leads to be the guardians of the Made Tech Delivery Standards, ensuring the team understands and is proactively practising them. A Delivery Lead is expected to encourage conversation on both more widely evolving the Delivery Standards, and ahead of deciding to follow a different path on their delivery.
+
+### Ensure at all times that the customer would recommend Made Tech to their peers
+
+We expect our Delivery Leads to take accountability for ensuring the customer is delighted with the quality of our delivery to the level that they would not hesitate to recommend Made Tech to their peers.
```
jonellison: Delighted's a bit of an airy term to judge performance against isn't it? Nobody's ever said: "I'm delighted that you've refactored all of that legacy code"

roles/delivery_manager.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Provide accurate and honest reporting of Delivery Healthcheck to Directors
+
+We expect our Delivery Leads to proactively provide regular reporting to the Directors on the health of their delivery.
+
+### Ensure that at all times the team is delivering according to the Made Tech Delivery Standards
+
+We expect our Delivery Leads to be the guardians of the Made Tech Delivery Standards, ensuring the team understands and is proactively practising them. A Delivery Lead is expected to encourage conversation on both more widely evolving the Delivery Standards, and ahead of deciding to follow a different path on their delivery.
+
+### Ensure at all times that the customer would recommend Made Tech to their peers
```
jonellison: Wouldn't ensuring this at _all_ times possibly stop you from having honest conversations?

If you've got the option to sweep something under the carpet to keep the client happy in the short term, or have a bit of short term dent in your relationship to avoid a bigger problem later on then the directive shouldn't be recommendation at all costs.

roles/delivery_manager.md
```diff
@@ -0,0 +1,63 @@
...
+
+We expect our Delivery Managers to proactively manage and mitigate risks to our deliveries and customer relationships. We expect our Delivery Managers to ensure the customer is delighted with what we deliver for their budget.
+
+### Define, capture and share necessary reporting materials
+
+We expect our Delivery Managers to work with customers to the define the frequency and type of reporting that will be most valuable to them.
+
+### Ensure teams and individuals take responsibility for delivery quality
+
+We expect our Delivery Managers to ensure teams are delivering with acceptable levels quality. We expect Delivery Managers to encourage their team to identify and proactively 'pull' from others when they need testing support. We expect Delivery Managers to *not* provide a safety net to their team by automatically testing features.
```
jonellison: > with acceptable levels quality

with acceptable levels _of_ quality

roles/delivery_manager.md
chrisblackburn: Oh, Jon. You and your good points.

jonellison: Nah, I'm just a pedant. I see where it's coming from after Xmas though - something about safeguarding reputation and client recommendations in the longterm maybe?

chrisblackburn: No, I think you're very right. The reality is that sometimes you need to cause some upset in the short term for the better long term outcome

PR #117, opened 2018-01-08
Add zendesk to services in new starter checklist

lukemorton:


chrisblackburn:
AFAIK we don't setup individual user accounts for Zendesk (@davidwinter?)

idabmat:
We use a shared account for Zendesk. Details are in LastPass

lukemorton:
Ah!

PR #118, opened 2018-01-10
Add Delivery Healthcheck

chrisblackburn:


team-norms/delivery_healthcheck.md
```diff
@@ -0,0 +1,17 @@
...
+ - When was your current iteration started?
+ - When is your next showcase?
+ - Are you on track to meet customer expectations?
+
+## Next iteration goal
+ - What is the goal for your next iteration?
+
+## Retrospective
+ - When was your last retrospective?
+ - What were your last retrospective action points?
```
danielburnley: more a general question - what if you've not had a retro since the previous one. Repeat the same action points or just say like "We've not had one since" ?

team-norms/delivery_healthcheck.md
```diff
@@ -0,0 +1,17 @@
+# Delivery Healthcheck
+
+It is the responsibility of the lead on each delivery team to provide an update on the health of their delivery in each week's BizOps showcase.
+
+A Delivery Healthcheck should answer:
+
+## Current iteration goal
```
kjdchapman: Is "What is your current goal" one of the things to share? If so might want to pull this out of the header to be clearer, and re-head as just "Current iteration".

team-norms/delivery_healthcheck.md
```diff
@@ -0,0 +1,17 @@
+# Delivery Healthcheck
+
+It is the responsibility of the lead on each delivery team to provide an update on the health of their delivery in each week's BizOps showcase.
```
zuren: Something on why?

team-norms/delivery_healthcheck.md
kjdchapman: All the other questions are pretty punchy - this is one that could go long, so worth keeping an eye on.

If retros only come out with a couple of action points, likely not a problem. If they have 5+ then this could take a while!

chrisblackburn: Yes, I'd expect exactly that. Of course, if you come to more than a couple of BizOps showcases without having had a more recent retrospective, questions may be asked as to your decision to eschew more frequent retros

chrisblackburn: Yep - let's see how this one pans out over the first few weeks and adjust accordingly

kjdchapman:
How many bonus points for acronyms on shaky ground?

**M**ission
**I**teration start date
**N**ext showcase
on-**T**rack?

nex**T** mission
**E**arlier retrospective
**A**ctions from retrospective

MINT TEA

PR #123, opened 2018-01-16
Clarify "billable" expectation in engineer spec

kjdchapman:
As Luke was running me through the expectations, the original version wasn't clear to me - not being familiar with working primarily with external clients.

Hopefully this is a bit clearer.

### DIFF
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -54,7 +54,7 @@ We expect our Engineers to be able to work on entire features, from conception t
 
 ### To be fully billable
 
-We expect our Engineers to be fully billed on customers deliveries as we are a business trying to make a profit by delivering valuable working software to our customers.
+Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Madetech isn't a sustainable business.
 
 ## Core Engineer Expectations
 
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -54,7 +54,7 @@ We expect our Engineers to be able to work on entire features, from conception t
 
 ### To be fully billable
 
-Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Made Tech isn't a sustainable business.
+We expect our Engineers to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Made Tech isn't a sustainable business.
 
 ## Core Engineer Expectations
 
--- a/roles/engineer.md
+++ b/roles/engineer.md
@@ -54,7 +54,7 @@ We expect our Engineers to be able to work on entire features, from conception t
 
 ### To be fully billable
 
-Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Madetech isn't a sustainable business.
+Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Made Tech isn't a sustainable business.
 
 ## Core Engineer Expectations
 
--- a/roles/as_code/lib/engineer.rb
+++ b/roles/as_code/lib/engineer.rb
@@ -54,7 +54,7 @@ JobSpec::Role.definition 'Engineer' do
     'We expect our Engineers to be able to work on entire features, from conception through to production deployment, without guidance and direction from others. This doesn\'t mean they have to be delivered without pairing, or asking advice from the wider team. What it does mean is the primary force for ensuring the feature is built and deployed in a timely fashion is the responsibility of the Engineer working on the problem.'
 
   expected 'to be fully billable',
-    'We expect our Engineers to be fully billed on customers deliveries as we are a business trying to make a profit by delivering valuable working software to our customers.'
+    'We expect our Engineers to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Made Tech isn\'t a sustainable business.'
 
   include CoreEngineerExpectations, as: 'Core Engineer Expectations'
 end


roles/engineer.md
```diff
@@ -54,7 +54,7 @@ We expect our Engineers to be able to work on entire features, from conception t
 
 ### To be fully billable
 
-We expect our Engineers to be fully billed on customers deliveries as we are a business trying to make a profit by delivering valuable working software to our customers.
+Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Madetech isn't a sustainable business.
```
craigjbass: `Made Tech`

craigjbass: What does `significantly contribute to customer deliveries` actually look like?

roles/engineer.md
```diff
@@ -54,7 +54,7 @@ We expect our Engineers to be able to work on entire features, from conception t
 
 ### To be fully billable
 
-We expect our Engineers to be fully billed on customers deliveries as we are a business trying to make a profit by delivering valuable working software to our customers.
+Engineers need to significantly contribute to customer deliveries, to justify billing their time. Without billing customers, Made Tech isn't a sustainable business.
```
craigjbass: What does `significantly contribute to customer deliveries` actually look like?

roles/engineer.md
kjdchapman: Thanks, corrected the name.

roles/engineer.md
craigjbass: How does one `justify billing their time`

kjdchapman: Could drop to just "contribute" - trying to imply the bar is "we can justify billing", which is how I understood this.

craigjbass: I think this has to tie into delivery standards + core skills

kjdchapman: I'm not sure what you mean - the core skills that are just below this section?

We don't mention delivery standards at all in this spec (yet), but I think that should be a new section rather than part of "billable".

I don't feel like I have a deep understanding of this section yet - I just wanted to clarify the language a bit. If it's not clearer in this version, happy to take suggestions.

craigjbass: I think "justify billing" is open-ended and potentially subjective(?)

The word "justify" also has quite invasive connotations in my mind which I'm not sure if it's exactly the spirit of what we're trying to do.  

In theory, if our https://github.com/madetech/handbook/blob/master/team-norms/delivery_standards.md are met, and an Engineer has technical skill as defined by our core skills, then you are fully billable. @chrisblackburn can you think of any situation where this would not be the case?

kjdchapman: I'd certainly be happy to drop the section - I think you're right that it's sufficiently implied by the technical skills and delivery standards.

If that was the case, I'd like to add some specific reference to the delivery standards here?

craigjbass: I'd like to see crystal clear clarity on the definition of "fully billable"

chrisblackburn: AFAIK an important commercial justification for us in hiring at engineer level is that we can, in good faith, fully bill the role. I'm not currently in favour of removing this requirement.

Core skills being attained are a likely barometer for being fully billable. That said, I expect there are a number of cases where we're currently billing for engineers where there is some additional work required to reach our core skills (from my current minds eye view of what the core skills will look like). 

So I'm not sure there's a 1:1 correlation here. I'd also question the timeframes on attaining the core skills. I don't necessarily expect we'll be asking new engineers to attain them within the first two weeks. Which is still the timeframe we're looking for from a billing perspective.

kjdchapman: (it's dawning on me I might not be the right person to write this, given my newness)

If I understand right, the expectations are useful to us because:
- They give us something to make hiring decisions around
- They help us to understand the performance of existing engineers (as well as helping those engineers improve)

If these are the case, what does this expectation add that isn't already covered through skills?

Is there a situation where we would hire someone because we see their have the necessary skills, but we can't see them being "billable"? What would that look like?

Is there a situation where we want to help someone with performance because they're not "billable", and it's not through improving specific, individual skills?

craigjbass: > (it's dawning on me I might not be the right person to write this, given my newness)

I suspect you are exactly the right person to be questioning/exploring this

chrisblackburn: > Is there a situation where we would hire someone because we see their have the necessary skills, but we can't see them being "billable"? What would that look like?

I'd think more typically the flip. A very common example is someone who shows strong engineering / TDD chops, but doesn't currently have the frontend skills. Likely we could hire an individual and soon have them in a billed capacity, but perhaps they're a way away from achieving core skills.

kjdchapman: As I think about this more, I do think either removing the section or moving it elsewhere is my recommendation. Note that this isn't a challenge to its importance (we do need to make money!), but rather the practical benefit of including this expectation here. 

If it's acceptable to hire someone who doesn't meet all of these expectations, that could be expressed in the guides to hiring.

If it's necessary for someone in this role to work towards being "billable", does that mean anything outside of "work on these other expectations"? If not, then it's not adding any extra information.

lukemorton: I think Chris is trying to make the point that the billable expectation needs to be met ahead of the other expectations.

It wouldn't be acceptable for someone in the Engineer role to be non-billable for longer than the probationary period. Other expectations on the other hand would be acceptable on a case by case basis.

kjdchapman: Can you help me to understand what justifies billing an engineer to a customer? Is it based on their ability to contribute to delivery?

I think Craig's concern about my original draft being woolly on the language is a valid one.

(I'm still unsure about the benefit of including this here, but I'm happy to continue revising instead of removing if people feel strongly about its inclusion.)

lukemorton:
If we are changing wording can we change them in the source files found in `roles/as_code/` and then use the `Makefile` to build the markdown? :)

kjdchapman:
@lukemorton done, thanks.

So, I want to get this either merged or closed. I'm in favour of removing the section entirely, but I think @RoryMacDonald, @chrisblackburn and @lukemorton are all keen to keep it. I'm not sure if @craigjbass  was in favour of removal or just further clarification.

Since it looks like the majority are in favour, I plan to merge this at EOD, in the absence of further objection. It might not be perfect, but I think it is an improvement.

chrisblackburn:
I prefer the wording suggested in this PR.

@craigjbass and I have something booked for tomorrow afternoon to see if we can land on a more concrete definition of what fully billable equates to - historically it's been a bit of "you know it when you see it".

kjdchapman:
@chrisblackburn No objection to me merging this before that discussion?

chrisblackburn:
Nope, I think this is an improvement as-is.

PR #126, opened 2018-01-17
Add security to core engineering expectations

lukemorton:
We are adding security expectations to all roles to ensure everyone has followed the security guidelines we've defined for keeping our devices secure. This hasn't been made clear to new starters in the past but want to ensure going forward everyone is made aware of what they are personally responsible for in terms of the security of their devices.


RyanMacG:
>You won't be required to enter verification codes when logging in with Google, which has its own 2-Step verification system.

From the trello 2FA page

lukemorton:
I'd suggest moving it to a mixin when it's needed and not before.

If there wording is all good I can merge now and change when we add the delivery manager role in.

PR #127, opened 2018-01-23
Add expectation around core competencies to Engineer role

lukemorton:
We expect every Engineer to be working towards attaining our core skills.


chrisblackburn:
Perhaps there's an individual expectation we set on people, based on the level on which they join us, as to how soon we expect them to attain these expectations. I don't think it needs documenting here necessarily, just musing.

PR #128, opened 2018-01-23
Introduce Senior Engineer role expectations

lukemorton:
Introducing a Senior Engineer role, sharing some expectations with Lead and defining it's own too.

I'd be interested to know if anyone thinks we're missing any here.

### DIFF
--- a/roles/as_code/lib/lead_engineer.rb
+++ b/roles/as_code/lib/lead_engineer.rb
@@ -1,4 +1,5 @@
 require_relative './shared_expectations/core_engineer_expectations'
+require_relative './shared_expectations/senior_engineer_expectations'
 require_relative './shared_expectations/delivery_lead_expectations'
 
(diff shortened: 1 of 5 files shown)


chrisblackburn:
Maybe something about picking up new languages/frameworks/tools easily?
Championing / leading the rollout by example of new technologies (thinking Richard with Elixir on Evermore, for example)?
Expert lean/agile? Or is this just code?

craigjbass:
@chrisblackburn Lean/Agile encourages behaviour where tech is used to further the team's efficiency. If we leave it out then I think these expectations massively fall down - "Tech for the sake of it"

chrisblackburn:
@craigjbass agree. It's not in there as an expectation for this role at present (AFAICS)

lukemorton:
@craigjbass and @chrisblackburn I believe I've addressed your suggestions.

PR #131, opened 2018-01-23
Add pension scheme to handbook

lukemorton:
We should probably tell people about our pension scheme :D


benefits/pension_scheme.md
```diff
@@ -0,0 +1,15 @@
...
+
+Below are the guidelines to the contributions we pay, and the minimum contribution you can pay.
+
+When | We pay | You pay | Total minimum contribution
+-----|--------|---------|----------------------------
+Now | 1% | 1% | 2%
+From April 2018 | 2% | 3% | 5%
+From April 2019 | 3% | 5% | 8%
+
+You can choose to increase your pension contribution. You can do this by speaking with a director or our finance department.
```
kjdchapman: Is it worth specifying how to contact the Finance department? (I think it's just finance@madetech.com, but I feel like I gleaned that magically from somewhere rather than reading it).

chrisblackburn:
https://github.com/madetech/handbook/blob/master/guides/policies/pensions.md

benefits/pension_scheme.md
chrisblackburn: Would rather people just spoke to finance direct also?

chrisblackburn: i.e. I can't add much value beyond saying "good idea, invest as much as you can in a pension"

lukemorton:
@chrisblackburn Have merged the two now Chris. Woops.

PR #134, opened 2018-02-01
Add "Account Lead" role

chrisblackburn:


roles/account_lead.md
```diff
@@ -0,0 +1,17 @@
+# Account Lead
+
+Account Leads are responsible for maintaining senior relationships within our customer organisations. They understand a customer's strategic objectives, and help teams navigate politically tricky situations.
+
+## Expectations
+
+### Point of escalation
```
lukemorton: I think Account Lead role is a point of escalation above the Delivery Director.

chrisblackburn: From the customer perspective I'd expect there to be one escalation point

Perhaps need clarity on team vs. customer escalations

PR #138, opened 2018-02-26
[WIP] Guidelines for choosing 'non-standard' technologies

chrisblackburn:


team-norms/README.md
```diff
@@ -21,3 +21,7 @@ Our way of soliciting feedback from and with our peers.
 ## [Retrospectives and Comradrospectives](retrospectives.md)
 
 Our methods of running team-wide and company-wide retrospectives.
+
+## [Supported Software](supported_software.md)
+
+Our list of software that is can be supported by the Support Team.
```
davidwinter: ~can be~

team-norms/supported_software.md
```diff
@@ -0,0 +1,25 @@
...
+The decision to self-host a third party application must involve the support team. You can expect to be encouraged to seek SaaS alternatives.
+
+Where the application is transitioning to a customer support team for ongoing support and maintenance, you're strongly advised to consult with the customer team on any major technology decisions.
+
+## Currently Supported Software
+
+- Ruby on Rails
+- Terraform
+- Ansible
+- MySQL
```
davidwinter: Wonder if we want to be more specific and say these should be AWS RDS variations only. No one wants to manage a DB these days...

chrisblackburn: Is this already covered earlier? "The decision to self-host a third party application must involve the support team. You can expect to be encouraged to seek SaaS alternatives."

team-norms/supported_software.md
```diff
@@ -0,0 +1,25 @@
+# Supported Software
+
+Where the day-to-day running of an application needs to transition to our support team, there are specific considerations you must take when making technology choices.
+
+As an organisation we're keen to embrace new and diverse technology where there's a strong commercial justification.
+
+When you're considering a major technology choice that isn't documented here, before starting a production implementation you must seek agreement from the support team that they're able to provide BAU support.
```
craigjbass: I personally think this is going to discourage teams from breaking from the list. While I agree that our software has to be supportable, I can see this being a barrier to entry to "diverse technology".

team-norms/supported_software.md
```diff
@@ -0,0 +1,25 @@
...
+
+The decision to self-host a third party application must involve the support team. You can expect to be encouraged to seek SaaS alternatives.
+
+Where the application is transitioning to a customer support team for ongoing support and maintenance, you're strongly advised to consult with the customer team on any major technology decisions.
+
+## Currently Supported Software
+
+- Ruby on Rails
```
craigjbass: Something doesn't feel right about this. Software isn't automatically maintainable because it uses ruby on rails.

chrisblackburn: Sure, there's more to maintainable software than technology choice, but what's being described here are technologies the support team are currently capable in offering BAU support for. A team decides C# is the the right tool for their job? They need to have a conversation with the support team before making that decision as it's placing an additional burden on the support team that they'll need to plan for.

chrisblackburn: You may consider this PR if you're looking at suggesting other capabilities software must provide to be maintainable: https://github.com/madetech/handbook/pull/139/files

team-norms/supported_software.md
chrisblackburn: I agree. I think there's a conflict between standardisation from a support perspective and desire to be more technology diverse. Suggestions for other ways to handle this conflict?

lukemorton:
Stale?

PR #139, opened 2018-02-28
Add BAU expectations

chrisblackburn:
A list of things an application and platform must provided to be considered ready for our Support Team to run

### DIFF
--- a/team-norms/bau_expectations.md
+++ b/team-norms/bau_expectations.md
@@ -6,7 +6,7 @@ The Made Tech Support Team requirements are identified below. Where the applicat
 
 ## Application Requirements
 
-- Centralised Logging (Loggly)
+- Centralised Logging (Papertrail)
 - Application Performance Monitoring (Datadog)
 - Exception Tracking (Sentry)
 - Automated Dependency Upgrades (Dependabot)
--- a/team-norms/bau_expectations.md
+++ b/team-norms/bau_expectations.md
@@ -1,0 +1,20 @@
+# Business As Usual (BAU) Expectations
+
+Teams must deliver software systems in accordance with our Business As Usual Expectations that make software systems easier to run, debug, and improve.
+
+The Made Tech Support Team requirements are identified below. Where the application is transitioning to a customer support team, you should liaise with the customer team to understand their specific needs.
+
+## Application Requirements
+
+- Centralised Logging (Loggly)
+- Application Performance Monitoring (Datadog)
+- Exception Tracking (Sentry)
+- Automated Dependency Upgrades (Dependabot)
+- Automated Security Patches (Snyk)
+
+## Platform Requirements
+
+- IaC provisioning (Ansible)
+- Server Monitoring (Datadog)
+- Automated Security Patching (operating system provided)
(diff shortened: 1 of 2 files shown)


team-norms/bau_expectations.md
```diff
@@ -0,0 +1,20 @@
+# Business As Usual (BAU) Expectations
+
+Teams must deliver software systems in accordance with our Business As Usual Expectations that make software systems easier to run, debug, and improve.
+
+The Made Tech Support Team requirements are identified below. Where the application is transitioning to a customer support team, you should liaise with the customer team to understand their specific needs.
+
+## Application Requirements
+
+- Centralised Logging (Papertrail)
+- Application Performance Monitoring (Datadog)
```
davidwinter: Unsure if we can fully go with this _currently_. I'm pretty sure when we started using Datadog for infra, the APM was free. Now they're charging for it too separate.

team-norms/bau_expectations.md
```diff
@@ -0,0 +1,20 @@
...
+- Exception Tracking (Sentry)
+- Automated Dependency Upgrades (Dependabot)
+- Automated Security Patches (Snyk)
```
davidwinter: Possibility we might be able to use Github security alerts for this now, which wouldn't require an additional charge. Needs testing though how good it actually is, if at all.

chrisblackburn: OK, what do you want to do in here? Leave blank to force conversation with team?

team-norms/bau_expectations.md
craigjbass: Yeah agree - we should get APM into projects as standard but all the options are fairly pricy.

team-norms/bau_expectations.md
```diff
@@ -0,0 +1,20 @@
...
+Teams must deliver software systems in accordance with our Business As Usual Expectations that make software systems easier to run, debug, and improve.
+
+The Made Tech Support Team requirements are identified below. Where the application is transitioning to a customer support team, you should liaise with the customer team to understand their specific needs.
+
+## Application Requirements
+
+- Centralised Logging (Papertrail)
+- Application Performance Monitoring (Datadog)
+- Exception Tracking (Sentry)
+- Automated Dependency Upgrades (Dependabot)
```
craigjbass: We're using package manager installed gitlab (k8s helm chart). 

Downside is that dependency upgrades don't work with gitlab. 

Upside is we have a make task that upgrades dependencies in < 30 seconds and tests that are safe enough to rely upon.

Is Dependabot a requirement?

team-norms/bau_expectations.md
chrisblackburn: @davidwinter did we figure out New Relic was going to do the job we need on the free tier for now?

team-norms/bau_expectations.md
chrisblackburn: I'd say everything here is a requirement unless you can reach agreement with the support team that they're going to be able to support your own custom setup

lukemorton:
Stale?

PR #141, opened 2018-03-23
Add value of "Continuous Improvement"

lukemorton:
Whenever I describe the culture of Made Tech, I describe it with the value of Continuous Improvement. From peer based 121s, to Director 121s for career development, to team retrospectives and our company wide comradrospective. We empower ourselves to reflect and improve at every level.

I feel like Continuous Improvement is up there with People, Space and Culture. I hope everyone else agrees but I'd encourage you to disagree openly if you do!

The wording may not be quite right so I'd appreciate some feedback in making sure the statement is clearly understood.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This handbook should be the starting point for any new team members. It provides
 
 * **Culture:** We value a positive and diverse company culture, and will always strive to adapt our culture in order to make this a more welcoming place to be.
 
-* **Continuous Improvement:** We value continuous improvement, supporting it with a blame free culture, and seek to empower ourselves to be able to reflect and improve.
+* **Continuous Improvement:** We value continuous improvement, and seek to empower ourselves and the people around us with the opportunity to reflect and improve.
 
 ## Contributing to our handbook
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ This handbook should be the starting point for any new team members. It provides
 
 * **Culture:** We value a positive and diverse company culture, and will always strive to adapt our culture in order to make this a more welcoming place to be.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,8 @@ This handbook should be the starting point for any new team members. It provides
 
 * **Culture:** We value a positive and diverse company culture, and will always strive to adapt our culture in order to make this a more welcoming place to be.
 
+* **Continuous Improvement:** We value continuous improvement, supporting it with a blame free culture, and seek to empower ourselves to be able to reflect and improve.
```
sowasred2012: The "blame free culture" sets off an alarm for me, not sure why. The best way I can articulate how it makes me feel is: imagine going to a swimming pool and there was a sign saying "there's DEFINITELY no man eating sharks in here".  I don't think I'd go in!

Plus, I think that point is implicitly covered by the People value - having real respect for the people you work with should curb the rise of a potential blame culture.

Maybe "We value continuous improvement, and seek to empower ourselves and the people around us with the opportunity to reflect and improve."?

lukemorton: Yeah I see what you mean. Best to reaffirm the positive in the value statement rather than dwell on the negative. Thanks Scott!

RyanMacG:
I guess this isn't massively related but looking at this and the other values I feel we don't _always_ completely follow them. Would be good to have some agreement on how we should call that out beyond "We should just call it out when we see it".

lukemorton:
Yeah agree Ryan, comradrospective topic perhaps?

craigjbass:
"Continuous Improvement" is an emotionally draining activity. Especially since you will need to leave your ego at the door. I think that the magic sauce here isn't the continuous improvement, it's the environment where we support each other's ego and get each other through the difficult realisations so that we can improve.

lukemorton:
Agree, takes constant work. I still think the process of improving the self, the team and the company is a value we all represent.

PR #142, opened 2018-03-26
Add more items to Getting Started checklist

lukemorton:
There are some basic questions you should probably have answers for after onboarding about where office supplies can be found and how to get more.


lukemorton:
We'll be moving to maintaining a keyholder list based on seniority and need, with a list of names and telephone numbers in case of emergencies.

PR #146, opened 2018-04-10
Add explicit expectations between employees and the org

lukemorton:
We already list our values in the Handbook but I think it’s important for us to be explicit about the healthy tension between the expectations of employees and the organisation.

These expectations also set the context for escalation procedures for employees to raise an issue and also for the organisation to escalate issues with employees.

**Further Background**

One of our People objectives for Q2 (and Q1 but things got away from me...) is to provide an escalation procedure in the Handbook when expectations aren’t being met. We already have this for employees to [raise an issue](https://github.com/madetech/handbook/blob/master/guides/welfare/raising_an_issue.md), we are now looking at it from the other way when an employee isn’t meeting the expectations of their role.

A future PR will include the expectation health check linked to along with more formal capability procedures.

PR #148, opened 2018-04-11
Add Health Check and Escalation for Expectation Issues

lukemorton:
Continuing on from yesterdays PR #146 we add an Expectation Health Check document along with a Capability Procedure.

## Purpose of Expectation Health Check

The purpose of the Expectation Health Check is to provide a holistic feedback cycle to ensure we uncover any issues that may be causing someone to not meet the expectations of their role. We need to ensure they are fit and healthy, that they are getting the support they need, we identify clearly what needs to change and we put a plan in place for getting there.

## Purpose of Capability Procedure

The Capability Procedure is a standard HR document for escalating performance issues in the worst case which can lead to written warnings and dismissal. The hope is that the Expectation Health Check can resolve almost all issues however we should have a more formal process in place too.

As usual feedback welcome!

### DIFF
--- a/guides/process/people/capability_procedure.md
+++ b/guides/process/people/capability_procedure.md
@@ -1,0 +1,26 @@
+# Capability Procedure
+
+In cases where employees are not meeting the expectations that have been clearly defined to them, and after an expectation health check has been performed, we follow the [Acas Code of Practice on Discipline Procedures](http://www.acas.org.uk/publications).
+
+## Expectation Health Check
+
+Before we escalate to formal proceedings, an expectation health check will be performed by a Director. Capability Procedures usually include measures to identify and address issues, our expectation health check seeks to address issues before reaching that stage.
+
+In the case where a health check has failed to address issues raised, a formal performance review procedure will be followed.
+
+## Formal Performance Review
+
+- Step 1: If it is decided a formal performance review must be carried out, Made Tech will notify the employee in writing of the reasons why and be invited to a meeting to discuss the problem.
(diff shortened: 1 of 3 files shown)


guides/process/people/capability_procedure.md
```diff
@@ -0,0 +1,24 @@
...
+## Expectation Health Check
+
+Before we escalate to formal proceedings, an expectation health check will be performed by a Director. Capability Procedures usually include measures to identify and address issues, our expectation health check seeks to address issues before reaching that stage.
+
+In the case where a health check has failed to address issues raised, a formal performance review procedure will be followed.
+
+## Formal Performance Review
+
+- Step 1: If it is decided a formal performance review must be carried out, Made Tech will notify the employee in writing of the reasons why and be invited to a meeting to discuss the problem.
+- Step 2: Made Tech will then hold the meeting to discuss the problem, allowing the employee to present their case. The employee has the right to bring a companion* with them.
```
danielburnley: `*` ?

kjdchapman: Does this * indicate a footnote? I can't see one.

lukemorton: You're both right!

lukemorton: You're both right!

kjdchapman:
I like this for two reasons:
- It's clear about what all the people involved should do
- It comes from the angle of "help the person improve" first

chrisblackburn:
Can this be merged?

PR #149, opened 2018-04-17
Report non iteration based workflows

chrisblackburn:
Some teams are working in non-iteration based workflows, which is making reporting the state of their deliveries hard. This PR tries to provide alternative means for reporting which may better fit the state of a kanban delivery.

Teams remain guided to question the best means of shaping their deliveries, particularly observing the non-committal place pure kanban flows can leave us and our customers.

### DIFF
--- a/team-norms/delivery_healthcheck.md
+++ b/team-norms/delivery_healthcheck.md
@@ -6,14 +6,25 @@ It is the responsibility of the lead on each delivery team to provide an update 
 
 A Delivery Healthcheck should answer:
 
-## Current iteration
+## Where you're working in defined iterations
+
+### Current iteration
  - When was your current iteration started?
  - What is your current iteration goal?
  - When is your next showcase?
  - Are you on track to meet customer expectations?
 
-## Next iteration
+### Next iteration
  - What is the goal for your next iteration?
+
+## Where you're working in more of a kanban flow
+
+ - What business objective are you working towards?
+ - What is your current work in progress?
+ - What are you currently blocked on?
+ - What did your team achieve during the past week?
+
+### Always
 
 ## Retrospective
  - When was your last retrospective?
--- a/team-norms/delivery_healthcheck.md
+++ b/team-norms/delivery_healthcheck.md
@@ -17,15 +17,15 @@ A Delivery Healthcheck should answer:
 ### Next iteration
  - What is the goal for your next iteration?
 
-## Where you're working in more of a kanban flow
+## Where you're working in a kanban flow
 
  - What business objective are you working towards?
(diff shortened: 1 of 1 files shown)


team-norms/delivery_healthcheck.md
```diff
@@ -6,15 +6,27 @@ It is the responsibility of the lead on each delivery team to provide an update
...
  - What is your current iteration goal?
- - When is your next showcase?
- - Are you on track to meet customer expectations?
 
-## Next iteration
+### Next iteration
  - What is the goal for your next iteration?
 
-## Retrospective
+## Where you're working in a kanban flow
```
craigjbass: I'm wondering whether `defined iterations` vs `kanban flow` is the right wording here.

On one hand you have iterations which produce a released increment of working software, that the customer can use *now* **end-to-end** versus what we're seeing which is a released increment of working software that works within a particular boundary e.g. API exposed on microservice X, UI built against design against stubbed API, etc. 

I think that it's important to not use the wrong wording because I see us as working in a kanban-based iterative workflow on all our deliveries.

naliwajek: Major business goal stays the same over the long period of time and we define everything you listed for it: a timeframe (giant roadmap in Excel) and a collection of capabilities (Trello).

Then we pull minor goals and tasks according to whatever is happening - what is blocking us, what 3rd party is available for meeting with us, when is the next showcase at which business people require some update so they can fight for a budget and so on.

So kinda 1 and 2 together IMO.

kjdchapman: I feel similarly to Jakub - we're not cleanly in one of these two boxes. 

We do work towards a short-term goal (generally, "what do we want to show at the next showcase"?), and I think that's a healthy pressure for us.

We don't fully define the whole iteration though, rather pulling in small tasks one at a time from the backlog. You could argue that freedom within the iteration removes that "good pressure". We've had good feedback from showcases, though.

kjdchapman:
I'm happy with this format - if it's giving the recipients what they need, let's merge it.

kjdchapman:
@chrisblackburn Team Akzo-Green has been following this for a few weeks. Has it worked for you? 

If so let's merge it! 

If not, let's change it!

naliwajek:
Better than the previous questionnaire, 7/10.

naliwajek:
(Can elaborate later - now I'm doing this TGIF and some meetings setup 😉- but good enough for me to merge and replace the old one.)

kjdchapman:
@chrisblackburn did you close it because it isn't giving you what you need?

PR #150, opened 2018-04-28
Policy on customers and industries that make people uncomfortable

chrisblackburn:
…le uncomfortable


RoryMacDonald:
Content LGTM. I think we should add a link to it from the listing

PR #156, opened 2018-05-08
fix link to the BYOD policy in the welcome pack

cybojenix:


kjdchapman:
@cybojenix for simple changes to the handbook like this, you're welcome to merge them in after they've got at least one approval. The onus is on the PR opener to get 'em merged.

(If the change is more controversial, you might want an RFC first)

PR #157, opened 2018-05-14
Add Client Principal expectations

chrisblackburn:


RoryMacDonald:
I'm in two minds about this. I see the Client Principal role primarily being about account management and taking responsible for things like:

- Making sure customers are happy.
- Defining and owning an account plans.
- Finding ways to wider our impact within customer organisations
- Delivering X financial growth within accounts YoY
etc

The current job description feels too too delivery focused for me, though it makes sense within the context of what we've discussed recently.

lukemorton:
We discussed alternate titles like CTO / Tech Director / Engineering Manager. We ended up conflating the concepts of Client Principal and CTO / Tech Director / Delivery Director under the current title Chris and I have on engagements. I'm not stuck on the idea of keeping the role definitions in one, we could split them out. Not sure what we gain though.

PR #161, opened 2018-06-20
Use "career development 121" terminology

lukemorton:
Trying to settle on consistent language! I think this is it.


kjdchapman:
I like this.

To elaborate, I like focusing on what the meeting is _for_ rather than _who_ is in it. This is a key thing for good meetings.

PR #178, opened 2018-08-02
Add gravatar to services setup

StevenLeighton21:
- We use a few things that you can link to gravatar for a shared profile picture (slack, guthub, continuous feedback)

### DIFF
--- a/company/welcome_pack.md
+++ b/company/welcome_pack.md
@@ -46,7 +46,7 @@ Below is a checklist for you to work with your buddy over the coming weeks.
 #### Non-Engineers
 * [ ] Complete the [Github tutorial](https://guides.github.com/activities/hello-world/)
 
-Below you'll find a list of tools that you will need to do your job. If you haven't already been provided with login credentials, then please let [Rory](rory@madetech.com), [Chris](chris@madetech.com), [Luke](luke@madetech.com) or [Seb](seb@madetech.com) know. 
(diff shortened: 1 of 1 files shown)


company/welcome_pack.md
```diff
@@ -66,6 +66,7 @@ Note that Google office applications (Docs, Sheets) are the preferred format for
 * [ ] LastPass (with 2FA)
 * [ ] Trello (with 2FA unless using Google Auth to login to Trello)
 * [ ] Continuous Feedback
+* [ ] Setup an account with [Gravatar](https://en.gravatar.com/) - You can use this to add a shared profile picture for multiple services.
```
lukemorton: We should add something about avatars needing your face, ideally from a professional picture.

StevenLeighton21: Can you suggest wording? ` You should add a profile picture that is clearly you, ideally professional looking portrait.`?

SebAshton: Also, should we explicitly state they should be "G" rated?

cybojenix: Should a photo policy be in a separate place, in a different PR? we already have a requirement elsewhere to have a photo for the website

StevenLeighton21: Seems like rules you add after they're broken to me, I'd reasonably expect any new starter to know what would be suitable

PR #179, opened 2018-08-22
Remove Friday drinks from benefits

kjdchapman:
We talked about this at a recent Comradro. 

It wasn't clear to me if this is something that we've already done, but just not recorded in the handbook, or if's something that's still up for discussion.

In case it is still up for debate:
We want to include people who don't want to drink alcohol. If we spend money on something else instead (#2ndfridaylunchgroup), then it might be able to include more people.


lukemorton:
Alternative view: We could consider changing language to Friday Social. We still invite new joiners to Friday evenings as well as Friday lunches and I usually use the "social" wording rather than "drinks". Remember the wording we have in our hiring: "We are vegan and non-drinker friendly as well as meat-eater and drinker friendly."

If everyone feels like we don't need this benefit I'm not against removing it. I'd also be up for discussing what else we can invest in to encourage more socials.

kjdchapman:
@lukemorton 
> We are vegan and non-drinker friendly as well as meat-eater and drinker friendly.

This is good context, thanks.

> I'd also be up for discussing what else we can invest in to encourage more socials.

I'd also love to see this. Not "let's stop spending money on X", but "let's do Y _instead_".

It's possible this PR isn't complete until it has a replacement in this vein.

lukemorton:
Can I suggest we close this PR until we've decided on a replacement benefit?

kjdchapman:
Yes, though I'm not sure what the next steps are - or if I should even own this (though I'm certainly happy to).

I opened this on the thought that it had pretty much been decided, but given some conversations I've had since, it looks like it hasn't.

PR #181, opened 2018-08-24
Clarify office dress code

chrisblackburn:


guides/office/dress_code.md
```diff
@@ -0,0 +1,21 @@
...
+- T-shirts
+- Trainers
+- Flip flops
+- Smarter shorts
+- Hoodies
+
+
+## Generally not OK for the office:
+
+- Tracksuit bottoms / sweatpants
```
SebAshton: I also feel vests/singlets are inappropriate, can we add this?

danielburnley:
I think it's easier to focus on things we don't really allow than what we do - as I think that list is far smaller than the things we're okay with

chrisblackburn:
Yeah, I'm open to that @danielburnley. It was more meant to be a list of things that might be considered 'borderline' in some environments. e.g. people may often ask are jeans and trainers OK, whereas trousers and smart shoes may be widely considered acceptable in most organisations.

guides/office/dress_code.md
lukemorton: Good catch!

chrisblackburn:
Maybe that in and of itself could be clarified

guides/office/dress_code.md
chrisblackburn: I'm probably going to regret opening this can of worms, but I feel like vest tops can be considered relatively smart for people identifying as female, whereas for those identifying as males much less so.

chrisblackburn:
@EThom73 I worry without being explicit that we're leaving it to interpretation (which is kinda informally what we've done so far). We've had a couple of occasions of people at customer offices in sweatpants for example, which is the sort of situation we'd be looking to avoid here.

I wonder if there are other way we can provide very clear guidance?

guides/office/dress_code.md
```diff
@@ -0,0 +1,10 @@
+# Dress Code
+
+Made Tech is typically a casually dressed environment, however, we should keep in mind that we remain a professional organisation.
+
+This should be considered a rough guide as to acceptable dress in the Made Tech office. If you're working on-site or visiting a customer organisation you should be sensitive to their dress code and modify your dress accordingly. If in doubt please seek advice from another member of the team.
+
+## Generally not OK for the office:
+
+- Tracksuit bottoms / sweatpants
```
georgeschena: I personally don't agree with this, when you're sitting down for 6+ hours its not always comfortable, hence why somedays I wear black tracksuit bottoms. I fully understand that if I see a client or visit a clients offices I wear jeans, I did this when I went to LAA and I dressed appropriately, however in the office I would like to feel as comfortable as possible while I sit and code for hours. @chrisblackburn

lukemorton: Unfortunately we have customers visit unexpectedly or have to head to a meeting without much notice so we do need to create guidance based on that requirement.

lukemorton: @chrisblackburn Could probably add a bit more of an explanation as to why we need to have a dress code?

chrisblackburn: Added to the first paragraph on the need to project a level of professionalism at all times

PR #183, opened 2018-09-04
Update scheduling around Comradros and Retros

RyanMacG:
Unless things have changed very recently the guidance around when retros
and Comradrospectives should happen seemed out of date. This should
update it to be more inline with what was the norm while I was at Made
Tech.

### DIFF
--- a/team-norms/retrospectives.md
+++ b/team-norms/retrospectives.md
@@ -23,7 +23,7 @@ There are two elements to retrospectives:
 
 ### Schedule
 
-Retrospectives should occur at least once every two weeks and should take 30-60 minutes.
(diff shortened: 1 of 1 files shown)


team-norms/retrospectives.md
```diff
@@ -16,13 +16,16 @@ The prime directive should be read before a retrospective begins.
...
-These retrospectives occur every 2 weeks on a Friday. There are 2 hours booked for this purpose. 
-
 There are two elements to retrospectives:
 
 * Retrospective (smaller groups performing retrospectives)
 * Comradrospective (whole company retrospective format)
 
+### Schedule
+
+Retrospectives should occur at least once every two weeks and should take 30-60 minutes.
```
craigjbass: Academy retros have a 90 min time box but we rarely need the full 90 mins. #JustSayin'

PR #184, opened 2018-10-18
adding information about Christmas

elena-vi:


benefits/flexible_holiday.md
```diff
@@ -41,3 +41,8 @@ We want to ensure teams own their holidays and that includes thinking about the
 * Time off due to illness and/or injury are handled separately. Flexible holiday cannot be used for these types of absences.
 
 * If a team members performance declines significantly due to abuse of this policy, then we reserve the right to review their use of this policy.
+
+## Christmas at Made Tech 
+
+* Made Tech closes for Christmas and New Year - you will not need to book this time off.
+* This year (2018) Made Tech will be closed from the 24th of December, and coming back to work on the 2nd of January.
```
SebAshton: We should be explicit whether the office closure is inclusive or exclusive of the 24th

danielburnley: Will we need to update this for every year or can we be more general in the rules for when we close? (No idea if it's always a case by case basis)

elena-vi: So after talking to @lukemorton the system of deciding this is looking at the calendar week and going "errr, I guess we don't need to come in for a day" so maybe updating it every year isn't such a bad idea?

elena-vi: yes

lukemorton:
If we adjust this to be just general information about us closing every year on Christmas Eve and then reopening day after New Years Day I think that will suit every year.

cob16:
This pr cannot be reopened after the default branch migration so have been opened over here https://github.com/madetech/handbook/pull/517

PR #190, opened 2018-11-13
Copy across onboarding list for Engineers & tidy checklists

craigjbass:


company/welcome/other-roles.md
```diff
@@ -0,0 +1,9 @@
+# Non-Engineers
```
chrisblackburn: Looking at the content of this it looks like it's sales specific rather than all non-engineer roles

lukemorton:
I'd prefer to simplify the checklist and keep it all in one easy welcome pack. Yasemin is moving some of this checklist into Airtable which will remove the need to go into as much detail.

Let's check with Yas about what needs to remain and simplify rather than introduce multiple checklists.

PR #191, opened 2018-11-15
Add guidelines around 3rd party services

EThom73:
**Purpose:** To formalise the operational process around how we manage 3rd party services/SaaS subscriptions.

### DIFF
--- a/guides/3rd-Party-Services.md
+++ b/guides/3rd-Party-Services.md
@@ -1,0 +1,17 @@
+# Use of 3rd Party Services
+
+We use 3rd party services/SaaS subscriptions accross both delivery teams and in the wider business. The services we use range from monitoring sytems for customers to trello plugins to improve roadmapping.
+
+These 3rd party services sometimes involve paying on a company credit card or storing company or customer data. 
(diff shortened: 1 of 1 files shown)


guides/3rd-Party-Services.md
```diff
@@ -0,0 +1,17 @@
+# Use of 3rd Party Services
+
+We use 3rd party services/SaaS subscriptions accross both delivery teams and in the wider business. The services we use range from monitoring sytems for customers to trello plugins to improve roadmapping.
+
+These 3rd party services sometimes involve paying on a company credit card or storing company or customer data. 
+In an effort to ensure we are storing this information responsibly, if you wish to use a 3rd party service for any reason, you will need to submit a [**3rd Party Services** form](https://docs.google.com/forms/d/e/1FAIpQLSeAysZFwhSDHzYrrKr_2T3KzRhCwZ3klRiBylM6KETeSxY8NA/viewform?usp=sf_link)
```
DelphineGattaz: My only concern with this new process is the waiting time for validation/rejection of the requests. Can we ensure that an answer will be given reasonably rapidly? (say within one day?)

EThom73: Yes. in the email (which much of this is copy/pasted from) Chris said minutes for approval.

Basically enough time for us to check it doesn't exist in another form.

EThom73: Also I don't see many scenarios for rejection. Pretty sure everyone here is reasonable and can legitimately make a case for something that would be valuable.

PR #193, opened 2018-11-19
Update README - 3rd parties.md

EThom73:
Adding link to guide for 3rd party expenses


chrisblackburn:
Oh, probably not one for this PR, but we should probably rename the file to `3rd_party_services.md` to keep the convention used elsewhere in this repo

PR #207, opened 2018-12-19
Add timesheeting expectation

craigjbass:


roles/academy_engineer.md
```diff
@@ -56,4 +56,8 @@ We expect our engineering team to understand how to secure the devices they use
 
 ### To be involved in practice improvement discussions
 
-We expect our engineering team to reflect on our practices and be a proactive voice in suggesting change. We expect you to be reflecting on your own development practices, reflecting on your team and the wider business.
\ No newline at end of file
+We expect our engineering team to reflect on our practices and be a proactive voice in suggesting change. We expect you to be reflecting on your own development practices, reflecting on your team and the wider business.
+
+### To keep timesheets up-to-date and submitted before invoicing deadlines
```
adrianclay: I believe the expectation is that they're submitted before 13:30 every Friday.

kjdchapman: This might be too specific for a handbook note?

craigjbass: @EThom73 might want to share an update to this, I think we should leave it open

PR #208, opened 2019-01-02
Add Head of Delivery role

chrisblackburn:


roles/as_code/lib/delivery_manager.rb
```diff
@@ -62,23 +63,6 @@
...
-    'We expect our Delivery Managers to work with customers and Account Leads to the define the frequency and type of reporting that will be most valuable to them.'
-
-  expected 'facilitate dialogue with Account Lead',
-    'We expect our Delivery Managers to facilitate a regular dialogue with their Account Lead to ensure both parties are abreast of relevant account and delivery activity.'
-
-  expected 'ensure teams and individuals take responsibility for delivery quality',
-    'We expect our Delivery Managers to ensure teams are delivering with acceptable levels of quality. We expect Delivery Managers to encourage their team to identify and proactively \'pull\' from others when they need testing support. We expect Delivery Managers to *not* provide a safety net to their team by automatically testing features.'
-
+  include DeliveryManagerExpectations, as: 'Delivery Manager Expectations'
   include DeliveryLeadExpectations, as: 'Delivery Lead Expectations'
```
lukemorton: Should `DeliveryLead` be `EngagementLead`?

roles/as_code/lib/head_of_delivery.rb
```diff
@@ -0,0 +1,80 @@
...
+    We are a technology focussed business that values continuous improvement, experimentation, trusting individuals and teams to do the right thing.
+
+    We are a growing business and now is a good time to join us as there is great opportunity to have a massive impact on the business and grow with us too.
+
+    ## Salary
+
+    This role has a salary of £80-95k depending on experience.
+  DESCRIPTION
+
+  salary 80_000..95_000
```
lukemorton: Salary band should be progressive rather than overlap. Either reduce top end of Delivery Manager or increase this bottom end.

yaseminercan: The delivery manager range is £50 - £85. Would reduce top end to £80k there as its a fairly large range already.

May have to introduce a senior delivery manager level for career progression. Assume there will only be one 'head of delivery' required so there would otherwise be no promotion opportunity.

roles/as_code/lib/delivery_manager.rb
chrisblackburn: Pre-existing IP. Will happily approve a PR that changes this across the board :-)

PR #211, opened 2019-01-08
Update Delivery Manager salary to reflect Head of Delivery

chrisblackburn:
Likely a Senior Delivery Manager role to sit somewhere in the middle in future


yaseminercan:
Yes, that makes sense. Agree with the range.

PR #212, opened 2019-02-04
Update Academy Engineer expectations inline with v2

craigjbass:


craigjbass:
@lukemorton this is a good point - I will address this in a followup PR.

I've ordered them how I'd expect them to be checked off

PR #214, opened 2019-02-12
Creating and changing delivery teams process

EThom73:
**What:**
This is a guide to submitting a request to the Operations Team about starting a new delivery team for a new customer coming in from Sales, or changing the structure of an existing delivery team.

**Why we want to introduce this process:**
Spinning up a delivery team support is a process that requires clarity over the details before it can take place. These processes aim to enable the Operations Team to get new delivery teams started as efficiently and effectively as possible. It also ensures that the details of your request go to a consistent location and are not lost in slack direct messages.

### DIFF
--- a/guides/creating_and_changing_delivery_team.md
+++ b/guides/creating_and_changing_delivery_team.md
@@ -1,0 +1,16 @@
+# Creating and changing delivery teams
+
+## Purpose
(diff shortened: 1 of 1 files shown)


chrisblackburn:
Will be documented in Made Tech Way

PR #219, opened 2019-03-07
Update minimum salary of Lead Engineer

lukemorton:
Before Christmas we updated structure to reflect Academy => Engineer => Senior => Lead progression. We now update salaries to match progression.


yaseminercan:
👍

PR #223, opened 2019-03-19
Remove support as we no longer provide this service

chrisblackburn:


cybojenix:
Will this also need the customer support role to be removed?

chrisblackburn:
Accepting PRs @cybojenix

PR #228, opened 2019-04-08
Single task for DT specific infrastructure

sarrionandia:
Replace engineer onboarding tasks with specific task.

### DIFF
--- a/company/welcome_pack.md
+++ b/company/welcome_pack.md
@@ -81,9 +81,7 @@ Academy Engineers: you can skip these two for now, but you will probably need th
 
 #### Engineers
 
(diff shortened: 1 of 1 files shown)


chrisblackburn:
Do teams no longer use The Mole for accessing IP restricted services more widely? (I don't actually know the answer to this)

sarrionandia:
> Do teams no longer use The Mole for accessing IP restricted services more widely? (I don't actually know the answer to this)

I did onboarding today, and for this step was told it isn't necessary since my first project won't require it. This at least means it isn't universal.

lukemorton:
Looks good to merge @sarrionandia

PR #231, opened 2019-04-13
Add new page about keeping the toilets clean

heidar:
I just want to make it clear that this is not a big problem, and this prevents it from becoming one. :)

As the team grows, it becomes more of a social responsiblity to keep the
toilets clean.

This page sets some guidelines on some steps we can all take to keep the
toilets nice and clean.

### DIFF
--- a/guides/office/toilet.md
+++ b/guides/office/toilet.md
@@ -17,3 +17,13 @@ When using the toilets, please make sure you perform the following:
  - Keep paper towels and toilet paper off the floor
  - Put toilet paper in the toilet and paper towels in the bin. Do not flush paper towels
  - Consider making use of air freshener after using the toilet if you think it is needed
+
+### Wash your hands, your life depends on it
+
+Germs can surive on bathroom surfaces for a long time. Human feces can carry nasty things like E. coli, Shigella, Streptococcus, hepatitis A and E, norovirus, and others.
+
+Only around two thirds of people actually wash their hands after using the toilet. Think about that next time you're pair programming.
+
+To keep it simple, follow the rule of "If you entered the toilet, wash your hands". Rinsing with water is not enough. Use soap, and lather it properly, then rinse it off.
+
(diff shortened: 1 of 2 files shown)


chrisblackburn:
I'm unsure if this is a feature creep, but it could be interesting to expand this to include post-bathroom hand washing tips. It's a shared responsibility to prevent the spread of hand-foot-mouth, hep. A etc.

heidar:
> I'm unsure if this is a feature creep, but it could be interesting to expand this to include post-bathroom hand washing tips. It's a shared responsibility to prevent the spread of hand-foot-mouth, hep. A etc.

An excellent idea. I have added a section for this.

guides/office/toilet.md
```diff
@@ -0,0 +1,29 @@
...
+
+No one is too good to put in some hard manual labour and clean the toilets when necessary. Let's keep the toilets sparkling clean, together.
+
+## Toilet tenet
+
+When using the toilets, please make sure you perform the following:
+
+ - Flush the toilet after use, half flush if you can to save water (and the planet)
+ - If any fecal matter remains inside the toilet bowl, please use the toilet brush to clean it
+ - If the fecal matter is outside the toilet bowl, please seek medical attention
```
DelphineGattaz: This probably can't stay in here as it is.

DelphineGattaz:
I've made one comment, I think one line should be removed. The rest looks good to me!

guides/office/toilet.md
yaseminercan: Agree, need to rewrite.  do we need toilet guides?

yaseminercan: Can we remove the toilet guide please? :) I'll call it out on slack (if I have to). Unless it becomes a problem I would rather not have this guide.

lukemorton:
Closed at request of Yas.

craigjbass:
🤣

PR #233, opened 2019-05-16
Update security guide

C-gyorfi:
Update the security guide and add password policy

### DIFF
--- a/guides/security/protect_the_company.md
+++ b/guides/security/protect_the_company.md
@@ -8,7 +8,7 @@ When setting up new laptops, please ensure you follow the below guidelines and u
 
 ### General
 
-- [ ] Make a note of serial numbers, model information
+- [ ] Register the make, model and serial numbers information on Charlie HR
 - [ ] Two Factor Authentication and strong passwords for:
   - [ ] Last Pass
   - [ ] Gmail
--- a/guides/security/protect_the_company.md
+++ b/guides/security/protect_the_company.md
@@ -9,7 +9,7 @@ When setting up new laptops, please ensure you follow the below guidelines and u
 ### General
 
 - [ ] Register the make, model and serial numbers information on Charlie HR
-- [ ] Passwords must follow the [password policy](guides/security/password_policy.md)
+- [ ] Passwords must follow the [password policy](password_policy.md)
 - [ ] Two Factor Authentication and strong passwords for:
   - [ ] Last Pass
   - [ ] Gmail
--- a/guides/security/protect_the_company.md
+++ b/guides/security/protect_the_company.md
@@ -9,6 +9,7 @@ When setting up new laptops, please ensure you follow the below guidelines and u
 ### General
 
 - [ ] Register the make, model and serial numbers information on Charlie HR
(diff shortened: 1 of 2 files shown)


guides/security/protect_the_company.md
```diff
@@ -2,21 +2,22 @@
 
 Keeping laptops and phones secure is vitally important. We're a software company and many of us have access to secure systems (both our own and our customers).
 
-When setting up new laptops, please ensure you follow the below guidelines and update the [security information spreadsheet](https://docs.google.com/a/madebymade.co.uk/spreadsheets/d/1jn1OeUd2DVWJPlzf_ZkdjtNpXWuy0r4_0dxXBHuWeG0/edit?usp=sharing)
```
C-gyorfi: I believe this sheet is no longer in use?

cob16:
These guidelines roughly follow the NIST security guidelines. Can I suggest that we reference them as our guide or in case someone wants additional information?

link: https://pages.nist.gov/800-63-3/sp800-63b.html#memsecret

C-gyorfi:
> These guidelines roughly follow the NIST security guidelines. Can I suggest that we reference them as our guide or in case someone wants additional information?
> 
> link: https://pages.nist.gov/800-63-3/sp800-63b.html#memsecret
> 
> There is one difference however NIST recommends not to have complexity requirements as they do not force users to make more secure passwords
> 
> Quoted is the relevint extract from there research (source: https://pages.nist.gov/800-63-3/sp800-63b.html#appA).
> 
> > As noted above, composition rules are commonly used in an attempt to increase the difficulty of guessing user-chosen passwords. Research has shown, however, that users respond in very predictable ways to the requirements imposed by composition rules [Policies]. For example, a user that might have chosen “password” as their password would be relatively likely to choose “Password1” if required to include an uppercase letter and a number, or “Password1!” if a symbol is also required.
> 
> > Users also express frustration when attempts to create complex passwords are rejected by online services. Many services reject passwords with spaces and various special characters. In some cases, the special characters that are not accepted might be an effort to avoid attacks like SQL injection that depend on those characters. But a properly hashed password would not be sent intact to a database in any case, so such precautions are unnecessary. Users should also be able to include space characters to allow the use of phrases. Spaces themselves, however, add little to the complexity of passwords and may introduce usability issues (e.g., the undetected use of two spaces rather than one), so it may be beneficial to remove repeated spaces in typed passwords prior to verification.
> 
> > Users’ password choices are very predictable, so attackers are likely to guess passwords that have been successful in the past. These include dictionary words and passwords from previous breaches, such as the “Password1!” example above. For this reason, it is recommended that passwords chosen by users be compared against a “black list” of unacceptable passwords. This list should include passwords from previous breach corpuses, dictionary words, and specific words (such as the name of the service itself) that users are likely to choose. Since user choice of passwords will also be governed by a minimum length requirement, this dictionary need only include entries meeting that requirement.
> 
> > Highly complex memorized secrets introduce a new potential vulnerability: they are less likely to be memorable, and it is more likely that they will be written down or stored electronically in an unsafe manner. While these practices are not necessarily vulnerable, statistically some methods of recording such secrets will be. This is an additional motivation not to require excessively long or complex memorized secrets.

PR #237, opened 2019-06-10
Number of people away on a team as a fraction

chrisblackburn:
Clarify the number of people who should be away from a team at any one time as a fraction to account for larger team sizes

### DIFF
--- a/benefits/flexible_holiday.md
+++ b/benefits/flexible_holiday.md
@@ -22,7 +22,7 @@ If you wish to book holiday outside of these guidelines, please raise this reque
 
 Having several team members away at the same time can make life really difficult for everyone else. Try to forward-plan your holiday, so it doesn't have a detrimental affect on your team members or on the company.
 
-Our guideline on this is that only 1 person from each team may be away at any one time. If you wish to book holiday outside of these guidelines, please raise this request with a director.
+Our guideline on this is that only 1 in 3 people on each team may be away at any one time. If you wish to book holiday outside of these guidelines, please raise this request with a Director.
 
 ## Who approves holiday?
 


sarrionandia:
I'm wondering where this number came from, it seems like there are so many variables that issuing global guidance in terms of a number instead of in terms of a set of criteria sets us up for bad decisions. The criteria I'd use to make this decision are:

- Impact on capacity and therefore delivery dates
- Budget (is spend too high? too low?)
- Time of year, I'm happier if it goes up during the peak holiday times (bank holidays, christmas, etc.) because lack of collaboration means that less gets done anyway
- Client relationship, do they think we are not getting enough done?
- Who the person is, do they have someone they can delegate to?

Basically, there are lots of real world criteria for doing this and I can't imagine how I would ever take the 1/3rd number into account if I was putting my client and team first.

yaseminercan:
I agree, there are many variables that can influence such a decision. 

Why are we reviewing this? Has the team collective decision on time off stopped working?

Should it be a judgement call from Tech lead?

chrisblackburn:
You make some interesting points @sarrionandia. In answer to where this came from: a guess as to what may be typically reasonable. I think we observed that 1 person per team feels generally unreasonable / unworkable on some of our now larger teams.

Who do you think should be involved in making this decision? i.e. who do you think would typically have context on this criteria and be well placed to judge the impact?

chrisblackburn:
@yaseminercan this is just updating the number of people on a team who should be away at any one time from 1 to 1 in 3 to account for larger team sizes (e.g. 1 person away at any one time on a 10 person team not realistic)

sarrionandia:
> You make some interesting points @sarrionandia. In answer to where this came from: a guess as to what may be typically reasonable. I think we observed that 1 person per team feels generally unreasonable / unworkable on some of our now larger teams.
> 
> Who do you think should be involved in making this decision? i.e. who do you think would typically have context on this criteria and be well placed to judge the impact?

The person managing the day to day relationship with the client, such as a DM or tech lead. In a small team, maybe just a quick huddle decision. I think it makes sense to lay down some guidelines, and maybe a sense of "try very hard to find a way to say yes", maybe some example scenarios for small and large teams?

PR #246, opened 2019-07-24
Guides for connecting to VPN.

yusufsheiqh:
## What

- Add instructions on how to connect to the VPN for Linux and MacOS devices

## Why

- We need to know how to connect to the VPN!


chrisblackburn:
LGTM. The only question I have is whether we should provide guidance on what the VPN should and should not be used for?

e.g. 
Yes: connecting to IP restricted services, adding a layer of protection when working from unsecured cafe wifi
No: streaming videos

cob16:
@chrisblackburn will add something around that

PR #248, opened 2019-08-09
Last day checklist - more SAAS and 1password

sarrionandia:


guides/security/last_day.md
```diff
@@ -1,10 +1,14 @@
...
   - [ ] GitHub
-  - [ ] LastPass
+  - [ ] 1Password
   - [ ] Google
   - [ ] Slack
   - [ ] Trello
   - [ ] Harvest
+  - [ ] Recruitee
+  - [ ] Airtable Tables
+  - [ ] Small Improvements
```
yusufsheiqh: Not sure why AWS is on its own.

```suggestion
  - [ ] Small Improvements
  - [ ] AWS
```

guides/security/last_day.md
```diff
@@ -1,10 +1,14 @@
...
+
 - [ ] Revoke access to any VPNs
 - [ ] Delete AWS account
```
yusufsheiqh:
```suggestion
```

guides/security/last_day.md
```diff
@@ -1,10 +1,14 @@
...
 - [ ] Access to these accounts should be revoked upon completion of the engagement:
   - [ ] GitHub
-  - [ ] LastPass
+  - [ ] 1Password
   - [ ] Google
   - [ ] Slack
   - [ ] Trello
   - [ ] Harvest
+  - [ ] Recruitee
+  - [ ] Airtable Tables
```
yusufsheiqh:
```suggestion
  - [ ] Airtable
```

sarrionandia: You can't revoke someones airtable account because we don't have any sort of centralised access, you can only revoke access to individual tables

cob16:
Should there be mention of this being done for customers accounts too?

sarrionandia:
> Should there be mention of this being done for customers accounts too?

This should happen as soon as somebody leaves an account, not on their lasy day of Made Tech. Also, I think that is a discussion for a new PR.

cob16:
Could also add Heroku to the list

guides/security/last_day.md
yusufsheiqh: That makes sense. Does revoking access to G Suite automatically revoke access to Airtable?

sarrionandia: Only if you signed up with Google SSO

PR #249, opened 2019-08-09
Create ims.md

EThom73:
**What:**
IMS stands for Integrated Management System. It is how we manage Security and Quality via ISO 27001 and ISO 9001 (forthcoming)

**Why is it going into the handbook?**
In line with implementing our learnings and policies from the ISO 27001 certification process, we are updating our Security policies and documenting them. 
We are also getting ready to implement ISO 9001, so a section has been reserved for documentation relating to this.

### DIFF
--- a/ims.md
+++ b/ims.md
@@ -1,0 +1,43 @@
+# IMS - What is it?
+
+IMS stands for **Integrated Management System**. It combines all the overarching systems that Made Tech use to govern our business operations. More specifically, our IMS currently manages our Security and Quality systems which have been implemented as part of our work towards our certification in ISO 27001 and ISO9001. 
+
+## ISO 27001
+### What is ISO 27001?
+ISO 27001 is an internationally recognised standard in Security Management. 
+### Why is Made Tech implementing ISO 27001?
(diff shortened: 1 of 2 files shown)


ims.md
```diff
@@ -0,0 +1,43 @@
...
+### What is ISO 9001
+ISO 9001 is an internationally recognised standard in Quality Management
+### Why is Made Tech implementing ISO 9001
+Goods and Services can have varying degrees of quality. Made Tech's mission is "Improving software delivery in every organisation", and a large part of that is endeavouring to deliver high quality products that both meet our clients expectations, and meet the needs of real users. The policies implemented in line with ISO 9001 help us to acheive that goal. 
+
+The relevant people in the business are
+- Quality Officer: Chris Blackburn/Luke Morton (TBD)
+- IMS Manager: Emily Thomas
+
+# Guide
```
chrisblackburn: I think this should be a level 2 header, and then level 3 headers underneath it?

ims.md
```diff
@@ -0,0 +1,43 @@
+# IMS - What is it?
```
chrisblackburn: Nitpicking, but maybe this should be:

# IMS

## What is it?

PR #255, opened 2019-08-15
Update README.md

EThom73:
**What is the change?**
Changing the section previously labelled `Security` to `IMS`. This new file will contain all details of our Security and Quality policies

**Why are we making the change?**
This reflects our move to an IMS (integrated management system) for Security and Quality. This stems from our implementation of practices and policies in line with ISO 27001, for which we are certified, and ISO 9001, which we will soon be undergoing certification for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,12 +90,6 @@ Contributing to the handbook isn't only possible for `git` users, find steps to 
 * [Tap Guide](guides/office/taps.md)
 * [Dress Code](guides/office/dress_code.md)
 
-#### Security
+#### Integrated Management System (IMS)
 
-* [Acceptable Usage Policy](guides/security/acceptable_usage_policy.md)
-* [BYOD Policy](guides/security/byod.md)
-* [Laptop & Phone Security](guides/security/protect_the_company.md)
-* [Device profiles](guides/security/device_profiles.md)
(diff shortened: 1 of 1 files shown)


chrisblackburn:
I think the approach you suggest is more user friendly. I think something like:

# Security
- What is IMS and ISO27001? (link to IMS overview)
- Existing links
- Links to new context excluding: classification and control, access control policy, information handling, outsourcing and supplier management

# Quality (I really hope there's a better header than this!)
- Future relevant ISO9001 content

EThom73:
An IMS  *can*  contain all overarching business operations, not just those limited to Security and Quality (as well as who knows what other certifications we will go for in the future). 

As "Security" breaks down into a number of components (physical security, byod, acceptable use, and more), the existing links do not currently cover all policies included under "Security". To me, it would give the impression that the policies listed on the handbook `README.md` are in some way more important than the others in the `IMS` section - which I would argue is inaccurate. 

Listing them here, and then listing them again on the `README.md` also seems to be an unnecessary reduplication of the IMS TOC , which will have to be updated every time we add a new document.

lukemorton:
The audience of the README doesn't need to know about IMS. They need to know about topics like Security, Quality and any other areas. We should focus on user needs, not organisational needs as we do in all our work. Let me know if you want a further discussion.

PR #258, opened 2019-08-15
Create classification_and_control.md

EThom73:
**What?**
Made Tech policy around the classification and control of assets and information

**Why?**
ISO 27001 compliance

### DIFF
--- a/ims/company/classification_and_control.md
+++ b/ims/company/classification_and_control.md
@@ -67,5 +67,5 @@ Public documents are meant to be in the public domain and have no access control
 Public information does not need to be labelled and has no handling or communication restrictions imposed.
 
 # Related Standards, Policies and Processes
- - [AUP Policy](aup.md)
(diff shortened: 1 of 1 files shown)


ims/company/classification_and_control.md
```diff
@@ -0,0 +1,71 @@
...
+ - Small Improvements public feedback
+ - Line management information in Charlie HR
+
+### Public
+Public documents are meant to be in the public domain and have no access controls. This information will not harm the organisation if released into the public domain, such as marketing material and public accounts.
+ 
+Public information does not need to be labelled and has no handling or communication restrictions imposed.
+
+# Related Standards, Policies and Processes
+ - [Acceptable Use Policy](aup.md)
```
lukemorton: I like that this is the full title. We should do this consistently. Equally we should use full name in filenames too.

PR #263, opened 2019-08-15
Create screening.md

EThom73:
What?
Made Tech's screening process

Why?
ISO 27001 compliance


EThom73:
I have asked Yas to update and expand the current hiring section (as it isn't in date).

IIRC we do need to specifically touch on security and hiring. If this can all be incorporated into that i think (🤞) we should be good.

lukemorton:
Cool, should we close this PR for now?

PR #265, opened 2019-08-15
Create onboarding.md

EThom73:
What?
Made Tech onboarding policy

Why?
ISO 27001 compliance


EThom73:
@lukemorton I *think* we need something like this to cover us from a documentation perspective of  "what we do", though it doesn't reflect the full checklist of actions to be taken. 

Could make it a bit more expansive on the "why" behind onboarding? e.g.

- Ensuring new MT Team members are confident in what is expected of them
- Providing the opportunity for a new starter to ask questions and discuss processes and policies with a human, rather than just being confronted with documentation
- a proper warm welcome to MT

lukemorton:
I think Yas just updated some of this checklist content. Probably worth looping her in to help align it.

PR #273, opened 2019-08-15
Create separation_of_development_test...

EThom73:
What?
Made Tech policy around separation of environments

Why?
Migrating from google doc for ISO 27001 compliance

### DIFF
--- a/ims/development/separation_of_development_test_production.md
+++ b/ims/development/separation_of_development_test_production.md
@@ -1,0 +1,8 @@
+# Separation of Development, Test, and Production environments
+
+This policy acts as a guide to our development practices and the separation of separation of development, Test, and production environments
+
+## The Policy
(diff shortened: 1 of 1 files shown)


ims/development/separation_of_development_test_production.md
```diff
@@ -0,0 +1,8 @@
+# Separation of Development, Test, and Production environments
+
+This policy acts as a guide to our development practices and the separation of separation of development, Test, and production environments
+
+## The Policy
+Separate, segregated, testing and development networks and systems are created on the appropriate cloud platform for the client's needs to provide safe environments for systems to be built and changes can be evaluated and tested with no danger of impacting production systems.
```
cob16: In my understanding, this policy is requiring a network-level separation of environments (e.g VLAN/firewall etc)? I just want to make sure that was the intention. (:pager: @lukemorton )

lukemorton: I think this is just a lightweight attempt to show we are using multi-environments rather than being policy that dictates what tech does. Technology department can elaborate on this in future to be more meaningful.

PR #287, opened 2019-09-05
Create onboarding.md

EThom73:
**What:**
An update on our onboarding process that covers who owns the responsibility for onboarding a new starter, what the process will involve, and how we security check all team members. 

**Why:**
Process not clearly outlined in handbook. There used to be a `first_day.md` file but it cannot be located (although it is referenced in some links).
We also need to provide this information in a clear, easy to access place for potential new employees, as well as for those going through the onboarding process so they know what to expect. 
This piece used to be part of a PR around IMS and was discussed with Yas (Head of people). It has now been edited and moved to the `Hiring` section as a more natural location for this information.

### DIFF
--- a/guides/hiring/onboarding.md
+++ b/guides/hiring/onboarding.md
@@ -1,0 +1,23 @@
+# Onboarding 
+
+The Head of People conducts an onboarding session with all new starters. This includes:
+ - Issuing the new starter with their device (unless they have opted for BYOD)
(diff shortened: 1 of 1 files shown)


lukemorton:
I think you were looking for `welcome_pack.md`: https://github.com/madetech/handbook/blob/master/company/welcome_pack.md

PR #288, opened 2019-09-05
Create Confidentiality_Agreements

EThom73:
**What:**
A short PR stating how Made Tech uses confidentiality agreements. 

**Why:**
It's very important to be clear and transparent around our use of confidentiality agreements. It not only ensures our team are handling information appropriately, but also instills confidence in our clients that Made tech is a safe and reliable partner to do business with. 

The use of NDAs for visitors also allows us to protect any sensitive work from falling into the wrong hands.

### DIFF
--- a/guides/security/confidentiality_agreements.md
+++ b/guides/security/confidentiality_agreements.md
@@ -1,13 +1,5 @@
-# Confidentiality Agreements
+# Confidentiality Agreements and NDAs
 
-At Made Tech:
- - All employees sign Terms and Conditions of Employment which include clauses for 
-   - Confidentiality
-   - Intellectual property
- - Company Non-Disclosure Agreements (NDA’s) are included in contracts raised with contractors 
- - Confidentiality requirements are also written into supplier contracts and agreements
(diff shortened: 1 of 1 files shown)


chrisblackburn:
Trying to figure out who the audience is for this one. Feels like maybe there's one thing to emerge:

- Non-disclosure agreements (people sometimes ask about how to do these - so guidance on that)

Could we lose the rest?

chrisblackburn:
And also add something helpful pointing to the NDA template etc.

EThom73:
OK. 

We don't seem to have a NDA template, but lots of versions. Is this the correct one? https://docs.google.com/document/d/1AcbQUpfiqgW_6ze0ewGrmUsIODnYRnzw1ZHxkw08w5Y/edit

chrisblackburn:
Yep, I think that's derived from the latest doc.

I think this is the template: https://docs.google.com/document/d/1YESByvaOv06azyq6JTNqmD9vsHh224lQ9urKV0p4_pY/edit

EThom73:
have updated the link and also tweaked the contents to cover situations where team members may need to refer to these guidelines

PR #289, opened 2019-09-05
Update raising_an_issue.md

EThom73:
Includes
1. Reference to head of People rather than Directors in process.
2. Outlines how Made Tech approaches disciplinary procedure. This is important for the purposes of transparency with our employees.

### DIFF
--- a/guides/welfare/raising_an_issue.md
+++ b/guides/welfare/raising_an_issue.md
@@ -39,7 +39,7 @@ Grievances are concerns, problems or complaints that employees raise with their 
 If a grievance cannot be settled informally, the employee should raise it
 formally with management, using the formal grievance procedure.
 
-- Step 1: Employee should write a letter or email to the directors of Made Tech Limited setting out the details of your grievance and how it might be resolved.
+- Step 1: Employee should write a letter or email to the Head of People setting out the details of your grievance and how it might be resolved.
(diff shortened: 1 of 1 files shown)


chrisblackburn:
As with the Capability Procedure (https://github.com/madetech/handbook/blob/master/guides/process/people/capability_procedure.md) perhaps instead of talking about "a process" we could mention that we follow ACAS CoP for dealing with disciplinary proceedings?

Otherwise LGTM, should probably have @yaseminercan give this a once over to merge too

EThom73:
Think that might be for a new PR? Don't know that it applies to the amendments I've made here.

chrisblackburn:
This PR adds "we have a disciplinary procedure in place" without providing information on that. I wonder if this is more genuine:

While we always prefer to handle minor misdemeanours in an informal way, we follow the ACAS CoP for handling disciplinary procedures, which will be invoked in the unusual event that a member of staff breaches any of our Company policies or performs (or assists in) an illegal or immoral activity.

yaseminercan:
Great to link to Acas. maybe we can add the link to acas too: https://www.acas.org.uk/article/2174/Discipline-and-grievance-Acas-Code-of-Practice

EThom73:
Cool. Have linked that directly.

chrisblackburn:
LGTM as long as @yaseminercan is happy

yaseminercan:
Happy :) all good from me

PR #290, opened 2019-09-05
Create office_security.md

EThom73:
**What:**
This is a breakdown of the information around the security and safety of our office. 

**Why:**
It details
1) That we have ensured our office buildings are safe for our team members and visitors. 
2) That we have guidelines to ensure visitors to the office don't see anything they shouldn't
3) The guidelines for the team on how to act if they suspect someone is trying or has tried to gain unauthorised access to our building

### DIFF
--- a/guides/security/office_security.md
+++ b/guides/security/office_security.md
@@ -1,0 +1,46 @@
+# Office Security
+
+Keeping our physical surroundings safe is an important part of ensuring our work remains secure. Robust office security ensures we maintain the safety of our buildings, team members, and visitors.
+
+## The Made Tech Head Office Details:
+- The Made Tech Head Office is in a multi-tenanted building located at 4 O’Meara Street, London SE1 1TE 
+- No environmental risks (such as flooding) have been observed
+- The landlord monitors the fire alarm and tests it weekly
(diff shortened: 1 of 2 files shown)


chrisblackburn:
I think this is trying to cover too much. Perhaps we need to rethink the use-case in which someone would be reading this guidance. E.g.:

- How to welcome visitors to the office
- Keeping the office secure
- Obtaining keys for the office
- I'm not currently sure of the role of the Head Office details

EThom73:
Propose following split of sections: 

1. Welcoming visitors section
2. General office security

IIRC The office details stuff is evidence that we have ensured the physical surroundings of the office have been assessed as posing no risk to employees/visitors/contractors

chrisblackburn:
I could see this being split in to separate guidance - the use case of wanting to get keys is quite different to welcoming guests to the office, for example.

Perhaps we can ask the question of whether we can get rid of this office details. I'm not currently sure what need this is meeting for anyone outside of an auditor.

EThom73:
OK

- have removed office details
- added mention of where to direct visitors
- have removed 'forced entry' section and will replace elsewhere

I can't find any section specifically around the getting keys piece.

PR #293, opened 2019-09-06
BYOD.md

EThom73:
Replacement for the old BYOD advice which got deleted. 
Language edited to make more appropriate


guides/security/BYOD.md
```diff
@@ -0,0 +1,101 @@
+# Bring Your Own Device Policy
+
+This document highlights things you should think about, and how you manage your own device while working on Made Tech products and services. Not only the data you work with but also the software you have installed and how you protect it - like antivirus software.
+
+This only applies to you if you’re using your own device for Made Tech work - including checking your Made Tech email. Devices aren’t just desktop and laptops - they include, for example, tablets and smartphones.
+
+## Using and managing your device
+
+If you’re using your own device for Made Tech work, it means you’re being trusted to use Made Tech information appropriately and keep any Made Tech information or data you hold, or can access, safe.
```
chrisblackburn: Feel like we can trim the last sentence without losing meaning to: and keep any Made Tech information safe

guides/security/BYOD.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Use endpoint protection software
+Unless your device is operating iOS or ChromeOS, your device must have endpoint protection software installed, which must also be kept up to date. 
+
+Endpoint protection is more than just antivirus software, and usually includes a firewall and may have other tools to protect your device, like network threat detection.
+
+Any device you plug into the device, like an external USB device, must also be scanned for viruses. We recommend doing a full scan at least once a month.
+
+### Set a password/passphrase
+Your device must require a password/passphrase to login, return from screensaver or wake from sleep. Ideally your device will lock the screen (or start the screensaver) after a few minutes (no more than 5) of the device being idle
```
chrisblackburn: I think we can tighten this up to: your device will lock the screen (or start the screensaver) after no more than 2 minutes of the device being idle.

guides/security/BYOD.md
```diff
@@ -0,0 +1,101 @@
...
+
+### How you manage passwords/passphrases
+Your password/passphrase must conform to the Made Tech Password Policy which will be linked below. You should also use a password manager and 2FA.
+
+### How you connect your device to networks
+We have our own VPN which automatically authenticates with G Suite. Everyone can use the service (not only engineers) for a more secure experience (i.e. protecting yourself when using a public WiFi). To connect to the VPN, please refer to the [guide found in the Handbook.](https://github.com/madetech/handbook/tree/master/guides/vpn) 
+
+### How you store & manage data
```
chrisblackburn: I'm not sure what this section achieves.

I think maybe we should rephrase this about being required to store data on backed up cloud services (G-Suite or GitHub)

chrisblackburn: Oh, that's already covered below. In which case I'm not too sure what this section covers

guides/security/BYOD.md
```diff
@@ -0,0 +1,101 @@
...
+If these aren’t encrypted, people can see any data that’s on them or install viruses without you noticing
+
+### Turn your device off, rather than putting it to sleep
+Depending on the device, data may only be encrypted when the device itself is turned off. So please be aware that closing the lid to put the device to sleep won’t necessarily enable encryption; in which case the data will not be protected.
+
+If your device will not be used for a significant period of time, we suggest turning it off entirely to ensure disk encryption is enabled. Most modern operating systems will allow you to resume all the files/applications that were open before shutdown.
+
+Turning off your device to ensure disk encryption is enabled is important if your device will be unattended. This could include when inside a backpack or carrying case when travelling, as those can be lost or stolen
+
+### Store data in Made Tech tools
```
chrisblackburn: That said this can probably be simplified to something along the lines of "Store data in G-Suite or GitHub. Do not store data locally on your device"?

guides/security/BYOD.md
EThom73: It was in the old approved PR so I copied it over. 

I'm not 💯 as to whether you are referring to the VPN snippet of the storing of data

guides/security/BYOD.md
EThom73: G-Suite or GitHub
- doesn't cover where all data is stored (Harvest/Xero/HubSpot/wherever the visitor ipad backs up to)

guides/security/BYOD.md
chrisblackburn: The 'store and manage data' section. I'm not convinced it's imparting any useful information. I'm not sure I'd consider do anything different after reading this, it's just telling me data can be lost or damaged.

guides/security/BYOD.md
chrisblackburn: As far as I can think G-Suite and GitHub are the only tools where you're likely to be working on a local copy of anything while needing to sync it back? And only G-Suite in quite extreme examples given Google Docs et al.

I think the ultimate guideline here is do not store data locally on your device, ensure it's synced to the relevant cloud service. Do we actually need to say any more than that?

StevenLeighton21:
Suggest wording instances around 'Made Tech data' could be 'company proprietary data'. I feel that's a better descriptor and less arguably vague.

guides/security/BYOD.md
EThom73: Have deleted this section

guides/security/BYOD.md
EThom73: amended, but have left the line about 

> If you are working offline and cannot access the cloud service

chrisblackburn:
LGTM. One small thing I'm sorry I didn't pick up on earlier, should we avoid the acronym in the filename, e.g. `bring-your-own-device.md` instead?

chrisblackburn:
Or `bring_your_own_device.md` as I think we use underscores in this repo 🤦‍♂

EThom73:
BYOD was the pre-ISO name for this, but have amended

PR #295, opened 2019-09-23
Update expenses.md

JulieLeahy:
Deleted credit card offer


davidwinter:
We should delete the file `company_credit_card.md` too if it hasn't been already.

PR #298, opened 2019-09-26
Create data_protection.md

EThom73:
What:
This details the process to follow if a client or employee makes a request in regards to their personal data. 

Why:
Data Access Requests and Data Deletion Requests require careful handling and this guide should ensure that team members can confidently handle any requests they receive.

### DIFF
--- a/guides/security/data_protection.md
+++ b/guides/security/data_protection.md
@@ -22,6 +22,6 @@ For Made Tech data (e.g. requests from Made Tech employees about personal data),
 
 ## Customer and Partner Data
 
-Where a customer or partner makes a request about their data, seek written approval from the that they have verified the request and authorise Made Tech to collate and release this data from systems we have access to. They may also need to provide proof of ID for the requester. 
+Where a customer or partner makes a request about their data, seek written approval from the that they have verified the request and authorise Made Tech to collate and release this data from systems we have access to. They will also need to provide proof of ID for the requester. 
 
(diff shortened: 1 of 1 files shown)


chrisblackburn:
I wonder if the tl;dr guidance for most of the team is:

- For Made Tech data forward the request to the ops team who will handle it thusly
- For Customer data seek written approval from the customer that they have verified the request and authorise Made Tech to collate and release this data from systems we have access to

guides/security/data_protection.md
```diff
@@ -0,0 +1,27 @@
...
+ - We will follow up and ask for information on the types of data in question (e.g. type and timeframe)
+ - We will process the request accordingly and respond to the data subject in line with the legislation. 
+ - We will inform the individual that a record of their basic information (e.g. name and company) may be kept with a note to not contact again.
+ - We may ask for input and/or provision of data from other teams to comply with the request. 
+
+## Customer and Partner Data
+
+Where a customer or partner makes a request about their data, seek written approval from the that they have verified the request and authorise Made Tech to collate and release this data from systems we have access to. They may also need to provide proof of ID for the requester. 
+
+Once this is done, refer the request to the Operations Team and we will process the request in line with the process outlined above.
```
chrisblackburn: I read this section slightly differently. To me this was if we're being asked to extract data from a customers system about one of their customers. e.g. a Teacher comes to DfE and asks for all the data the department have on them.

If this is the case I think the team will need to execute the request, but should seek reasonable certainty that the request has been validated by the customer.

PR #299, opened 2019-09-26
Create security_team.md

EThom73:
Who's who in the Security team, what they are responsible for, and how to report an event or incident.


guides/security/security_team_and_reporting.md
```diff
@@ -0,0 +1,34 @@
+# The Security Team and reporting incidents
+
+## Who is responsible for Security at Made Tech?
+
+Each team is responsible for ensuring that robust security practices are happening within their own team. This applies to delivery teams (Delivery Lead, Tech Lead, engineers etc.) and to business teams (Sales, People, etc.).
+
+In addition to this, we have a Security Team who are responsible for ensuring secure practices are adopted across the business and that we continue to meet the standards required to acheive our security certifications (ISO 27001, Cyber Essentials +, etc.)
```
chrisblackburn: do we need the etc.?

chrisblackburn:
I think there are two use-cases here:

- Security team responsibilities and objectives
- How to report and incident

Pondering if it's useful to provide examples of the sort of things to report. e.g. Dan highlighting the Slack data thing was really useful but not OBVIOUS

chrisblackburn:
I don't think my earlier comment was clear. Thoughts on splitting this in to two separate documents? I feel like the use-cases are sufficiently different

guides/security/security_team.md
```diff
@@ -0,0 +1,21 @@
+# The Security Team and reporting incidents
```
chrisblackburn: I think this isn't correct now?

EThom73: Good spot. Thanks

PR #305, opened 2019-10-04
Update README.md

EThom73:
Updating the readme to include TOC links to new Security related documentation.


chrisblackburn:
Reading the document titles in this context I wonder if some more friendly names (for the links and the documents) may be something like:

* Who is responsible for security at Made Tech?
* Bring Your Own Device (BYOD) Policy
* Standard laptop setup and approved software
* Data Protection Policy
* Access to Systems and Data Policy
* Employee Last Day

Also there's an inconsistent use of Title Case

PR #313, opened 2019-11-26
Update Engineer role

lukemorton:
## Why

The description of the job was fairly out of date, and not as inclusive as it could be.

## What

Craig, myself and wider tech team have reviewed and updated content.

### DIFF
--- a/roles/as_code/lib/engineer.rb
+++ b/roles/as_code/lib/engineer.rb
@@ -4,46 +4,33 @@ JobSpec::Role.definition 'Engineer' do
 
     ## What does the job entail?
 
-    We primarily write and deliver custom software to our customers. We build solutions in a variety of languages and platforms, historically Ruby on Rails, we tend to use React, JavaScript, Elixir, and Clojure these days. We will usually be responsible for setting up a customer’s infrastructure on AWS with tools like Ansible and Terraform though sometimes we opt for a Platform as a Service like Heroku. We make sure there is a Continuous Delivery pipeline running our automated test suites before deploying changes.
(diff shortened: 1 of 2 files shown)


roles/as_code/lib/engineer.rb
```diff
@@ -4,46 +4,33 @@
...
 
-    The software we build solves business problems. We’ve built software for karaoke, e-commerce, retail buying departments, accountants, photography studios, warehouses, supply chain companies, and ourselves. Typical projects will last 3-6 months, some customers work with us over longer periods but we like to mix up teams at around 6 months to keep things fresh.
+    Our teams have used Ruby with Rails and Sinatra, ES6 with React and Angular 2, C# with .NET Core. We don’t limit ourselves as a company and we expect all our Engineers to be keen on learning new technologies. Automation is important to our teams, so we make sure there is a CD pipeline set up to build, test, and release. Also, we will usually be responsible for setting up a customer’s infrastructure too! Such as on AWS, GCP or Azure using a tool like Terraform, though sometimes we opt for a Platform as a Service like Heroku.
```
chrisblackburn: Lots of abbreviations (CD, AWS, GCP etc.)

Prefer "We aren't limited to a small set of technologies" over "We don’t limit ourselves as a company"

Unsure of the need for an exclamation! on! setting! up! infrastructure!

chrisblackburn:
Were we going to abandon this Ruby generator as it makes it hard for non-engineering teams to update job descriptions?

chrisblackburn:
I see this happened in a subsequent PR. Sad times :-(

PR #317, opened 2019-11-26
Remove all role expectations from engineer roles

lukemorton:
## Why

We are rolling out SFIA to all engineers this quarter and so expectations are no longer required.

## What

Remove all expectations from engineer roles.


cob16:
Will SFIA be added to the handbook in its place?

lukemorton:
Interesting question @cob16, they probably should be. I think having explicit expectations in the Handbook to indicate what we expect are really important. The new SFIA expectations are wrapped up in Spreadsheets at the moment which needs untangling. 

The current expectations in this Handbook are out of date and I think I'd prefer to remove them rather than have incorrect information in here.

kjdchapman:
Is a placeholder about SFIA is better than leaving a blank?

> We use [SFIA](https://www.sfia-online.org/en) to explain the expectations of a role. 
We're hoping to add more detail about this soon.

(The SFIA official docs aren't particularly friendly, so not sure)

adrianclay:
I remember reading these and thinking they were really useful context when applying/deciding whether I should apply to MT.

StevenLeighton21:
SFIA doesn't entirely replace these, does it? We're still working to uphold these expectations afaik.

lukemorton:
SFIA largely does replace these expectations and we've promoted people even though these expectations have not been met at all, so they are fairly inaccurate.

We will aim to add examples based on SFIA later this year.

PR #321, opened 2020-01-08
Add Senior Eng with AWS

lukemorton:
## What

Adding a Senior Engineer job advert with the requirement of strong AWS experience. While this job advert has a specific need for strong AWS experience, it’s for our standard Senior Engineer role. We aren’t moving away from our polyglot / generalist approach.

## Why

We have a specific need in Bristol over the next two years for AWS skills and so need to advertise for that particular skillset.

Anyone we hire needs to be a full stack / polyglot software engineer, or at least want to be long term.


yaseminercan:
Looks great! Thanks

PR #345, opened 2020-02-03
Adding salary range for Client Partner

yaseminercan:


chrisblackburn:
I think these role expectations are _really_ out of date - aligned to when Luke, Rory, and I performed the role rather than any more recent expectations. I think better to delete this one

yaseminercan:
Delete role expectations or the role?

lukemorton:
@chrisblackburn ?

PR #352, opened 2020-02-12
Update parental leave to not just refer to newborns

davidwinter:
Tried to make this text clearer and inclusive also to adopted children being welcomed into a family.


yaseminercan:
looks good to me :)

PR #354, opened 2020-03-16
Fix spelling mistake in Timesheets.md

paperclipmonkey:
Spelling mistake from "completed" to "complete" so the sentence flows correctly


cob16:
seems the fork did not trigger the build so this can't be merged :cry:

lukemorton:
Thanks @paperclipmonkey for the PR!

@cob16 Yes, it's probably safe that PRs from outside MT can't run in our CI environment, however I'm not sure what workflow would then enable us to merge. Will force a merge but we need to look into this.

PR #355, opened 2020-03-16
Delete security_policy.md

EThom73:
No longer true.
The specific Github articles now reflect our security policy.


cob16:
did a rebase to get the build to pass

PR #362, opened 2020-06-08
Fix dead link to SFIA skill description

kjdchapman:
I noticed this when opening an unrelated PR, as it fails the build.

@lukemorton / @chrisblackburn do you know if this link is necessary/useful? Good to not rely on external links when we can due to link rot.


kjdchapman:
Hmm, this doesn't fix it - not sure how the dead link detector works, as this link works in the browser (and returns `200 OK`). For speed, I'm going to remove this link for now...

PR #363, opened 2020-06-15
Updating benefit

yaseminercan:
Adding that candidates are unable to ask for flexible working before they join us as we are hiring for full time roles only (or mostly) unless its advertised as flexible working possible

### DIFF
--- a/benefits/flexible_working.md
+++ b/benefits/flexible_working.md
@@ -4,6 +4,7 @@ For a variety of reasons, people may choose to work fewer days per week than the
 
 Wherever possible we aim to accommodate this. 
 Note: This benefit is only available to current team members as the majority of our open roles are advertised as full time.
+We'll always aim to support working patterns where we can, due to client considerations, we have an internal process for requesting a flexible working pattern, and we default roles to a standard pattern
 
 ## Some things to consider
 
--- a/benefits/flexible_working.md
+++ b/benefits/flexible_working.md
@@ -2,7 +2,8 @@
 
 For a variety of reasons, people may choose to work fewer days per week than the usual Monday - Friday.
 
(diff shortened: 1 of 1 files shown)


cob16:
My understanding is this means having a flexy start and finish times to accommodate for things such as child care, appointments etc. Not moving to part-time. The opposite would be having a strict start at 9-5 ect

benefits/flexible_working.md
```diff
@@ -2,7 +2,8 @@
 
 For a variety of reasons, people may choose to work fewer days per week than the usual Monday - Friday.
 
-Wherever possible we aim to accommodate this.
+Wherever possible we aim to accommodate this. 
+Note: This benefit is only available to current team members as the majority of our open roles are advertised as full time.
```
StevenLeighton21, line 6: I'm not sure what this means to me as an external reader - there are potential benefits I can't have? 

Is this a temporary state?

How does this affect say, caregiving - could I work full time but still do school runs?

yaseminercan: We currently only hire full time roles, so to a candidate applying, currently, flexible working is not available, i.e. less than 5 days per week.

The flexible working time benefit would allow people to do their school runs etc though so think that caregiving in many cases is possible.

StevenLeighton21: I would suggest rewording this section to reflect that we can be flexible on day to day timings to accommodate things like care, but not on number of days/larger flex like working 5 days in 4 etc.

yaseminercan: Thats the flexible working times section for day to day timings. This section is particular to less working days. 
Do you think I should make a link to that for working times?

PR #367, opened 2020-06-24
Update VPN connection details for Ubuntu

antonyoneill:
I had a couple of issues when adding the VPN profile on Ubuntu 18.04.

The external link had some typos in the package names which caused me to stop reading, where I then missed the settings for the algorithms. 

I wanted to make it a little clearer what needs to be done to set this up without relying on another site which may change the content in the future.

### DIFF
--- a/guides/vpn/README.md
+++ b/guides/vpn/README.md
@@ -37,24 +37,43 @@ That is it! You should now be connected to the `Made Tech - VPN`. If you are exp
 
 ### Ubuntu/Debian
 
-https://docs.foxpass.com/docs/foxpass-linux-l2tpipsec
+1. Install the L2TP module for network-manager & gnome extension if necessary
+    ```bash
+    sudo apt update
+    sudo apt install network-manager-l2tp
+
+    # If in a gnome environment
+    sudo apt install network-manager-l2tp-gnome
+    ```
+2. Follow the steps to add the VPN connection below.
+3. You must set the Phase1 and Phase2 algorithms in IPSec Settings:
+   1. Open the Made Tech VPN connection settings
+   2. Navigate to Identity tab
+   3. Open IPsec Settings
+   4. Under Advanced:
+      - Phase1 Algorithms `3des-sha1-modp2048`
+      - Phase2 Algorithms `3des-sha1`
+   5. Save, close
+4. Connect & enter password
 
 ### Fedora
 
 1. If you are using the GNOME version, first install the dependencies
+    ```bash
+    # for GNOME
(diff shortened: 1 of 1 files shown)


guides/vpn/README.md
```diff
@@ -37,39 +37,60 @@ That is it! You should now be connected to the `Made Tech - VPN`. If you are exp
...
+2. Follow the steps to add the VPN connection below.
+3. You must set the Phase1 and Phase2 algorithms in IPSec Settings:
+   1. Open the Made Tech VPN connection settings
+   2. Navigate to Identity tab
+   3. Open IPsec Settings
+   4. Under Advanced:
+      - Phase1 Algorithms `3des-sha1-modp2048`
+      - Phase2 Algorithms `3des-sha1`
+   5. Save, close
+4. Connect & enter password
```
yusufsheiqh, line 57: Is there a reason for adding all of this which mirrors the content in the link? Foxpass is what we are using, so to me, it makes more sense to link to their docs directly.

The only difference I can see here are the phase algorithms being switched around.

guides/vpn/README.md
```diff
@@ -37,39 +37,60 @@ That is it! You should now be connected to the `Made Tech - VPN`. If you are exp
...
-4. Click the `IPsec Settings..` button
+3. Click the `IPsec Settings..` button
 
 ![main_settings_screen](images/linux/ipsec.png)
 
 - Check the top checkbox and enter the shared secret (this will be provided to you)
 - Click **OK**
 
-5. Click the `PPP settings..` button
+**On Ubuntu you must set the Advanced Phase1 & Phase2 algorithms as above**
```
yusufsheiqh, line 91: Ubuntu has its own section above. This section is for Fedora, so this isn't needed(?).

cob16:
If we are committing to maintaining our own copy of the ubuntu guide then can we keep the link in case it falls out of date with the official Foxpass one (the VPN backend uses Foxpass so they may change their settings in a future update)

guides/vpn/README.md
```diff
@@ -37,39 +37,60 @@ That is it! You should now be connected to the `Made Tech - VPN`. If you are exp
 
 ### Ubuntu/Debian
 
-https://docs.foxpass.com/docs/foxpass-linux-l2tpipsec
```
cob16, line 40: If we are committing to maintaining our own copy of the ubuntu guide then can we keep the link in case it falls out of date with the official Foxpass one? The VPN backend uses Foxpass so they may change their settings in a future update.

antonyoneill:
Closing out for now as I think it has limited value.

PR #369, opened 2020-06-26
another update to parental leave

yaseminercan:
Adding that you have to be with Made Tech for 26 weeks to be eligible
also phasing return to work starting with 30 hours per week before going back to full time
plus making clear which parts are paid by Made Tech and which are via government payment


guides/welfare/parental_leave.md
```diff
@@ -1,17 +1,20 @@
 # Maternity, Paternity & Shared Parental Leave
 
 Whether because you or your partner is having a baby, if you're adopting a child or having a baby through a surrogacy arrangement, Made Tech encourages you to take leave so that you can welcome the new child (or children!) into your family.
-Our paid Parental leave is the same for all!
+Our paid Parental leave is the same for all!
```
two-first-names, line 4:
```suggestion
Our paid Parental Leave is the same for all! 
```

Should this be capitalised?

PR #375, opened 2020-07-22
Add a page for software engineer 2

StevenLeighton21:
The page is more or less a copy of Software Engineer 1 with a few tweaks.

Kept the same wording around hiring into these roles with minor tweaks to both to cover Academy -> E1 -> E2 typical progression.


lukemorton:
Thanks for this @StevenLeighton21! Can we also update the link in `roles/README.md` as part of this change?

PR #385, opened 2020-08-05
Add line management landing page

kjdchapman:
This is a draft version of guidance for line managers. @yaseminercan and I have started on a guide for probation first.

Looking for feedback on:
- Is there a better term than "direct report"? Or, like "line manager", is this the best, honest term?
- Is there a way to make [this line](https://github.com/madetech/handbook/pull/385/files#diff-3e28a56dfcb47e38286fdbd145da065bR34) less scary?
- Does anything here misrepresent our current process?
- Is anything here scary for a reader looking to join us?


guides/line-management/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## What do I need to do as a line manager?
+
+### One-to-one meetings
+Hold regular one-to-one meetings, at least once a month.  
+
+### Probation
+Help your direct reports to pass their probation period.  
+Work with the Market Principal to extend or fail probations, if necessary.
+
```
chrisblackburn, line 20: Throwing these out: 
- Provide a judgement on whether Made Tech is the right environment for their direct reports to succeed long-term
- Assert that their direct reports meet the minimum standards for their role

guides/line-management/probation.md
```diff
@@ -0,0 +1,11 @@
+# Probation at Made Tech
+
+When you join Made Tech, you'll be "on probation" for a short time.
+
+This means your notice period is initially shorter (how long?).
```
chrisblackburn, line 5: typically one week during probation

guides/line-management/README.md
lukemorton, line 20: There's something about failing probation is really a sign of poor hiring rather than poor line management (or even poor performance if the issue is more about core values alignment).

kjdchapman: I like this. I think I would sink this detail into the "more guidance on probation".

guides/line-management/README.md
```diff
@@ -0,0 +1,35 @@
+# Line management at Made Tech
```
StevenLeighton21, line 1: This has been floated around before but maybe here is a good place to discuss -
The wording 'line management' leaves a lot to be desired, and has a connotation of management style attached to it.

I personally prefer the wording along the lines of `Career Mentor`. I think this still describes the responsibilities well, while removing some of the more negative associations such as performance management. It's not like a career mentor would shy away from/not be doing _management_ in some cases, but language is important and I feel a conscious effort to move in this direction would help build trust and open communication by presenting the career mentor as an ally and friend.

kjdchapman: @sarrionandia said something that really resonated with me a while ago on this - there's an 'honesty' in the term 'line manager', as it's someone who's responsible for things associated with that role traditionally, including the more difficult bits.

I hear you, though - I don't think we do this typically at Made Tech, and we miss that in the name.

StevenLeighton21: I don't disagree! I wonder if there's a dishonesty in wrapping those responsibilities in wording like mentor?

Could that have an adverse effect when actually doing the difficult bits? I had assumed not, that it may make the difficult bits easier to start on.

guides/line-management/README.md
kjdchapman: I plan to sink these into the specific probation guidance - coming soon™

guides/line-management/README.md
yaseminercan: I think 'Career Mentor' is a great name but in conflict with some of the aspects of line management that are not purely coaching/mentoring. Such as performance management. I think it could be misleading to call this role Career Mentor but it certainly is a big part of that role.

Although I'm also not a fan of the term Line Manager :) 
I have heard the word 'Sponsor' before but that might be unclear. Maybe we can review the role name as a group later

kjdchapman:
I've no longer included an example page with this; I plan to do this in a follow-up PR.

guides/line-management/README.md
StevenLeighton21: I'd probably argue performance management is mentoring in a sense, but getting into semantics/obfuscating that probably isn't advisable

PR #387, opened 2020-08-14
Merging minor and major changes

sarrionandia:
Following some discussion about the community norms of how we actually go about approving changes, I'm opening this PR to formalise it.

### DIFF
--- a/guides/contributing_to_the_handbook.md
+++ b/guides/contributing_to_the_handbook.md
@@ -51,6 +51,9 @@ the channels more relevant to those people.
 4. Manage the conversation, seek approvals on GitHub, address feedback.
 5. When it feels like something close to consensus is established, merge.
 
+Consensus is subjective, and we can't make a rule that catches all of the important ways it can look.
+If you're unsure if you've got enough of it, reach out to the community on Slack.
+
 ### 2. Uncontroversial clarifications
 
 This includes spelling mistakes, poorly worded sections, and writing something down that we all are doing anyway,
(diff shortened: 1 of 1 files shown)


sarrionandia:
> I think 'when it feels like a consensus is reached' is still woolly and I'm sometimes still unsure about hittin' that merge button.
> 
> Perhaps something about how that's ok and it's on the raiser of the change to just go for it, and if it's really actually not agreed upon you'd soon be made aware?

I'll try to draft something, I like the sentiment of this.

guides/contributing_to_the_handbook.md
```diff
@@ -30,3 +30,33 @@ This guide aims to make it easier to submit changes to the Handbook without usin
...
+### 1. Significant changes or additions
+
+This includes changes to policies, changes to the way people are expected to work, changes that express a departure from a previous community norm, etc.
+
+1. Open pull request
+2. Shout about the pull request in all relevant Slack channels. This should include announcements, but if it relates to a particular part of Made Tech, also in
+the channels more relevant to those people.
+3. Allow enough time for people to review and comment, remembering that they probably have busy schedules. One to two weeks is probably sensible.
+4. Manage the conversation, seek approvals on GitHub, address feedback.
+5. When it feels like something close to consensus is established, merge.
```
craigjbass, line 52: Some policies in the handbook are would have an impact on function of the business. I think it makes sense to attempt to seek out impacted parts of the organisation and seek their advice on a change.

I say this because while I get the feeling that people would pick up on and comment on all PRs that impact them, I don't have any evidence to prove that this is true.

sarrionandia: I tried to capture that in point 2 -- do you have a way of making it more clear?

guides/contributing_to_the_handbook.md
```diff
@@ -30,3 +30,36 @@ This guide aims to make it easier to submit changes to the Handbook without usin
...
+
+Consensus is subjective, and we can't make a rule that catches all of the important ways it can look.
+If you're unsure if you've got enough of it, reach out to the community on Slack.
```
StevenLeighton21, line 55: LGTM

PR #390, opened 2020-08-19
Add description of how scheduling decisions are made

chrisblackburn:
Scheduling, the dark art that sits somewhere between chess, a jigsaw puzzle where you don't have all of the pieces, and something about trying to ensure context is shared amongst all of the players.

Really it's about facilitating conversations between the right people. This describes the different types of "need" the scheduling team typically supports, who's involved in servicing each of these, and what decision we're trying to make.

As we've recently regionalised, there's likely an over-reliance on Market Principals here. As we scale out markets we'll expect some of this stuff may be delegating to members of a local leadership team.

Oh yes, highly recommended viewing in rendered markdown rather than source (biiiig table).


kalliedgar:
**may be delegating** - in the section above - potential typo - should this be delegated?

chrisblackburn:
Yep, it should be. This is just in the narrative rather than in the handbook content so I'll go ahead and merge

PR #398, opened 2020-09-25
Update acceptable_use_policy.md

fdelin:
Suggest removing the paragraph on reporting the arrival of visitors from this section in the handbook which looks like it should be about Made Tech IT device or service usage only, and if needed, as a follow-on change, update the section on "Visitors and Office Security" accordingly.

### DIFF
--- a/guides/security/acceptable_use_policy.md
+++ b/guides/security/acceptable_use_policy.md
@@ -23,9 +23,6 @@ I will exercise good judgement at all times regarding the reasonableness of any 
 ## I will enable  2-factor authentication where this is available
 2-factor authentication (2FA) requires two forms of authentication, often a password and a code, sometimes sent via SMS or an application. It helps protect user accounts in the case that your password is somehow exposed.  
 
-## I will ensure I report the arrival of guests and visitors promptly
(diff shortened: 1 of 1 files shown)


cob16:
Moving this information to a more relevant file feels fine to me. Although I am not sure what the intent of this document is. 

By the looks of it, it is a centralised agreement of things to do. Other orgs do things like require a signature for things like this.

StevenLeighton21:
> Moving this information to a more relevant file feels fine to me. Although I am unsure given the possible intent of this document.
> 
> By the looks of it, it is a centralised agreement of things to do. Other orgs do things like require a signature for things like this. Removing this line entirely sounds a bit like taking this responsibility away. Something that might be required for some of Org certifications.
> 
> Perhaps someone with more insight can clarify

We include the information in the handbook as a guide/reference, the policies are covered in our annual training which is attendance recorded/signed for.  I commented on the other PR that the change looks good, I would like to allow policy holders to approve also :)

StevenLeighton21 (approved):
I think enough time has passed for review that this is good to go!

PR #400, opened 2020-10-07
Add information about other options for paid counselling

Ninamma:
On top of Employee Assistance, we also have the option of finding a therapist/counsellor ourselves, which Made Tech will help pay for - this PR adds this information to the handbook.

I've renamed the existing file, so that it's more generic: `paid_counselling` with `employee assistance` and `other options` sections, and updated links.


yaseminercan:
Looks good to me :)
Thanks for updating Ninamma!

PR #403, opened 2020-10-26
Fix salary format across tech roles

tmaritz:
I noticed that the format of the salary section in tech role descriptions was inconsistent:

The salary for this role is location dependant:
_London_: £xxxxx to £xxxxx _Manchester_: £xxxxx to £xxxxx _Bristol_: £xxxxx to £xxxxx

vs

Bristol: £xx,xxx
Manchester: £xx,xxx
London: £xx,xxx

I've updated these as follows:

The salary for this role is location dependant:
Bristol: £xx,xxx
Manchester: £xx,xxx
London: £xx,xxx

### DIFF
--- a/roles/lead_software_engineer.md
+++ b/roles/lead_software_engineer.md
@@ -91,6 +91,8 @@ Compensating you fairly:
 
 ## Salary
 
+The salary for this role is location dependant:
+
 - Bristol: £63,900-£81,000
 - Manchester: £63,900-£81,000
 - London: £71,000-£90,000
--- a/roles/principal_technologist.md
+++ b/roles/principal_technologist.md
@@ -76,6 +76,8 @@ Compensating you fairly:
 
 ## Salary
 
+The salary for this role is location dependant:
+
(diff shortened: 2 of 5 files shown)


roles/lead_software_engineer.md
```diff
@@ -91,6 +91,8 @@ Compensating you fairly:
 
 ## Salary
 
+The salary for this role is location dependant:
+
 - Bristol: £63,900-£81,000
 - Manchester: £63,900-£81,000
 - London: £71,000-£90,000
```
WillGibson, line 98: Could Bristol and Manchester be together where they are the same?
```suggestion
- Bristol & Manchester: £63,900-£81,000
- London: £71,000-£90,000
```

StevenLeighton21: No guarantee they'll stay the same, or that we won't add more locations. Seems reasonable to me to duplicate for clarity?

WillGibson (approved):

PR #404, opened 2020-10-27
Add line manager guidance for probation period

kjdchapman:
This adds minimal guidance about 'probation', aimed at line managers.

I'm especially interested in feedback from new line managers / people recently joining Made Tech, but I'm happy to hear anyone's thoughts.

You may want to view the [rendered markdown pages](https://github.com/madetech/handbook/blob/a93bd9aa10630863ac683527374e78679208acff/guides/line-management/probation.md) when reviewing.

### DIFF
--- a/guides/line-management/probation.md
+++ b/guides/line-management/probation.md
@@ -20,7 +20,7 @@ At the end of probation, pass or fail it, and explain what happens next.
 
 ## Explaining how to pass probation
 
-Talk about what good looks like for their role. [The handbook role pages](../roles) may help with this.
+Talk about what good looks like for their role. [The handbook role pages](../../roles/README.md) may help with this.
 It's good to spend the first two weeks talking about this.
 
 Explain that feedback is useful evidence on how they're performing.
--- a/guides/line-management/probation.md
+++ b/guides/line-management/probation.md
@@ -8,6 +8,11 @@ Rarely, you'll extend probation, or pass someone early.
 
 This guidance is written for line managers, but may be useful for people on probation too.
 
+- [What happens during probation?](#what-happens-during-probation)
+- [Explaining how to pass probation](#explaining-how-to-pass-probation)
+- [One-to-one meetings during probation](#one-to-one-meetings-during-probation)
+- [Finishing probation](#finishing-probation)
+
 ## What happens during probation?
 
 In the first week, explain what probation is, and how to pass it.
--- a/guides/line-management/probation.md
+++ b/guides/line-management/probation.md
@@ -1,0 +1,84 @@
+# Probation at Made Tech
+
+When someone joins Made Tech, they're "on probation" for a short time.
+This means their notice period is shorter - usually one week.
+
+At the end of probation, people either pass or fail. Most people pass.
+Rarely, you'll extend probation, or pass someone early.
+
+This guidance is written for line managers, but may be useful for people on probation too.
+
+## What happens during probation?
+
+In the first week, explain what probation is, and how to pass it.
+
+Each week, meet your direct one-to-one.
+Get feedback from their team, and reflect on it together.
+Talk often about whether they're on-track to pass.
+
+At the end of probation, pass or fail it, and explain what happens next.
+
+## Explaining how to pass probation
+
+Talk about what good looks like for their role. [The handbook role pages](../roles) may help with this.
+It's good to spend the first two weeks talking about this.
+
+Explain that feedback is useful evidence on how they're performing.
+
+Tell your direct to ask their team for feedback. You can talk to your direct's team to help get more feedback, if needed.
+
+### Worries about failing probation
+
+It can be stressful to be on probation.
+
+Explain early that:
+- Most people pass
+- Getting regular feedback means the result won't be a surprise
+- You will guide and support them
+
+## One-to-one meetings during probation
+
+We don't have guidance for one-to-ones in the handbook yet.
+
+In brief, ask your direct about their week:
+- How are they feeling?
+- What's gone well?
+- What's been difficult?
+- What have they learned?
+
+For probation, specifically:
+- Reflect on feedback together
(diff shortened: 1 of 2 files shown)


guides/line-management/probation.md
```diff
@@ -0,0 +1,87 @@
...
+Talk about what good looks like for their role. [The handbook role pages](../../roles/README.md) may help with this.
+It's good to spend the first two weeks talking about this.
+
+Explain that feedback is useful evidence on how they're performing.
+
+Tell your direct to ask their team for feedback. You can talk to your direct's team to help get more feedback, if needed.
+
+### Worries about failing probation
+
+It can be stressful to be on probation. Explain early that:
```
StevenLeighton21, line 37: Maybe also re-iterate that probation is also useful for the individual to determine if working here is right for them, and also that extending probation because there's a need to see some competencies that aren't easily showcased _right now_ is relatively common and not a negative reflection

guides/line-management/probation.md
```diff
@@ -0,0 +1,87 @@
...
+At the end of probation, pass or fail it, and explain what happens next.
+
+## Explaining how to pass probation
+
+Talk about what good looks like for their role. [The handbook role pages](../../roles/README.md) may help with this.
+It's good to spend the first two weeks talking about this.
+
+Explain that feedback is useful evidence on how they're performing.
+
+Tell your direct to ask their team for feedback. You can talk to your direct's team to help get more feedback, if needed.
```
WillGibson, line 33: "direct" reads a bit weird. "direct report" would be more conventional terminology.

yaseminercan: agree

guides/line-management/probation.md
kjdchapman: I agree with all of this - but I'm concerned about adding prose here. I'm trying to address two things:
- I'm responsible for welfare, and I want to highlight that people worry about probation
- I'm responsible for line management, and I have someone worried about probation - what should I do?

More broadly: I think communities/discussion are a better place for the more nuanced points, not this guidance.

(side note: My understanding is that extending probation is relatively rare (and this document describes it as such) - @yaseminercan might have better data on that if it's in doubt.)

guides/line-management/probation.md
kjdchapman: addressed in https://github.com/madetech/handbook/pull/404/commits/945a407ebf65a885e2429b01ed72126f2f5f6c58

guides/line-management/probation.md
StevenLeighton21: >and I have someone worried about probation - what should I do?
I think explaining the way probation also benefits the employee is part of that reassurance, I'd consider that something a line manager should highlight. I'm ok with the nuance of that being part of discussions rather than guidance though

guides/line-management/probation.md
```diff
@@ -0,0 +1,89 @@
...
+
+Each week, meet them one-to-one.
+Get feedback from their team, and reflect on it together.
+Talk often about whether they're on-track to pass.
+
+At the end of probation, pass or fail it, and explain what happens next.
+
+## Explaining how to pass probation
+
+Talk about what good looks like for your direct report's role. [The handbook role pages](../../roles/README.md) may help with this.
```
StevenLeighton21, line 28: I think this should link to `roles/sfia` - the core competencies and behaviours described there are IMO the most useful reference when trying to determine what being successful in that role will look like

guides/line-management/probation.md
```diff
@@ -0,0 +1,89 @@
...
+### Worries about failing probation
+
+It can be stressful to be on probation. Explain early that:
+- Most people pass
+- Getting regular feedback means the result won't be a surprise
+- You will guide and support them
+
+## One-to-one meetings during probation
+
+We don't have guidance for one-to-ones in the handbook yet. In brief, ask your direct report about their week:
```
StevenLeighton21, line 44: I think we do have guidance now, which could be linked to instead

WillGibson: As someone about to be doing some line managing for the first time here, linking to that would be very useful.

guides/line-management/probation.md
kjdchapman: I think SFIA without context is scary - linking to `roles/sfia` gives you a list of docs rather than a formatted README.

I see it as the job of the roles "landing page" to surface whatever is most important - if that's SFIA, sure, but I'd change that page, not this link.

guides/line-management/probation.md
kjdchapman: Great point, shows how old this PR is :grimacing:

guides/line-management/probation.md
kjdchapman: I am coming round to this being a good thing to include, but I haven't found a good one-line expression of it. I'll think about it and we can hopefully iterate on this page further.

guides/line-management/probation.md
kjdchapman: Addressed in https://github.com/madetech/handbook/pull/404/commits/075deb2c3790b7c20daeb9d04617dbe516dad78d (and then links hopefully fixed in https://github.com/madetech/handbook/pull/404/commits/18d56126d092f1b476c3886c915a0b9c4c6698a6 :grimacing:)

guides/line-management/probation.md
WillGibson: You could start [here](https://github.com/madetech/handbook/blob/master/roles/sfia/academy_software_engineer.md), but maybe an index page would be better. The pages themselves are good.

WillGibson: Or there is this: https://github.com/madetech/handbook/blob/master/roles/README.md#technology

kjdchapman: That roles page that you've linked second is where this ends up - without the #technology anchor tag, since probation period is across non-tech roles too (as far as I know!)

PR #405, opened 2020-10-28
Equality, Diversity & Inclusion Policy

WillGibson:
Amongst other reasons, we need a policy on this so that we can start to run periodical surveys to see what the sate of play is in terms of Equality, Diversity & Inclusion at Made Tech.

This is a living document, which we expect to evolve over time, created by @WillGibson, @chadtomkiss-madetech, @yaseminercan and Yasemin Furlonger with contributions from other including @kjdchapman and @Ninamma.

### DIFF
--- a/guides/equality-diversity-and-inclusion/README.md
+++ b/guides/equality-diversity-and-inclusion/README.md
@@ -8,5 +8,5 @@ We aim at all times to promote equality, diversity and inclusion in the workplac
 
 We recognise that we are enhanced by different backgrounds, experiences, views, beliefs, cultures and physical/neurological differences represented within our team. We embrace diversity in all of our activities and acknowledge that variety and difference are intrinsic to the success and future development of our business. 
 
-Read our [Equality, Diversity & Inclusion Policy](guides/equality-diversity-and-inclusion/README.md).
+Read our [Equality, Diversity & Inclusion Policy](policy.md).
 
--- a/guides/equality-diversity-and-inclusion/README.md
+++ b/guides/equality-diversity-and-inclusion/README.md
@@ -8,5 +8,5 @@ We aim at all times to promote equality, diversity and inclusion in the workplac
 
 We recognise that we are enhanced by different backgrounds, experiences, views, beliefs, cultures and physical/neurological differences represented within our team. We embrace diversity in all of our activities and acknowledge that variety and difference are intrinsic to the success and future development of our business. 
 
-Read our Equality, Diversity & Inclusion Policy.
+Read our [Equality, Diversity & Inclusion Policy](guides/equality-diversity-and-inclusion/README.md).
 
--- a/guides/equality-diversity-and-inclusion/README.md
+++ b/guides/equality-diversity-and-inclusion/README.md
@@ -1,0 +1,12 @@
+# Equality, Diversity & Inclusion at Made Tech
+
+We aim at all times to promote equality, diversity and inclusion in the workplace and to provide a working environment that is free from discrimination.
+
+- We want everyone to feel respected and valued
+- We want to help everyone meet their potential
+- We hire, coach and promote without discrimination
+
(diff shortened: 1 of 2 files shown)


guides/equality-diversity-and-inclusion/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+We aim at all times to promote equality, diversity and inclusion in the workplace and to provide a working environment that is free from discrimination.
+
+- We want everyone to feel respected and valued
+- We want to help everyone meet their potential
+- We hire, coach and promote without discrimination
+
+We recognise that we are enhanced by different backgrounds, experiences, views, beliefs, cultures and physical/neurological differences represented within our team. We embrace diversity in all of our activities and acknowledge that variety and difference are intrinsic to the success and future development of our business. 
+
+Read our Equality, Diversity & Inclusion Policy.
```
chadtomkiss-madetech, line 11: Markdown link here?

guides/equality-diversity-and-inclusion/policy.md
```diff
@@ -0,0 +1,89 @@
...
+
+- Age
+- Disability
+- Gender reassignment
+- Marital or civil partner status
+- Pregnancy or maternity
+- Race (including colour, nationality, ethnic or national origin)
+- Religion, religious belief or similar philosophical belief
+- Sex
+- Sexual orientation (ie homosexuality, bisexuality or heterosexuality)
```
jbevan4, line 23: Do we really need to list out all the references from the external source? Given that we aim to go above and beyond this, can we not just have the link to the external source and use the end of 25?

guides/equality-diversity-and-inclusion/policy.md
```diff
@@ -0,0 +1,89 @@
...
+
+This list doesn't cover everything that might disadvantage someone. We're interested in fairness for everyone, not just meeting our legal obligations.
```
jbevan4, line 25: 👍 🎉

guides/equality-diversity-and-inclusion/policy.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Diversity
+
+This is a broader concept that builds upon the progress made through equal opportunities. Everyone is different and diversity means recognising, respecting and valuing the differences we each bring to work.
+
+Equal opportunities and diversity work together by identifying and addressing any inequalities and barriers faced by people and by valuing, learning and benefiting from the diverse cultures in society and our staff.
+
+### Direct discrimination
+
+This means treating someone less favourably than you would treat others because of a Protected Characteristic. For example, rejecting a job applicant because they are Jewish, or not promoting someone because they are gay. This includes any less favourable treatment because you perceive a person to have a Protected Characteristic (even though they do not in fact have it), or because they associate with a person or group who has a Protected Characteristic.
```
jbevan4, line 37: Do we want to link to an external definition of this such as [this](https://www.gov.uk/discrimination-your-rights/how-you-can-be-discriminated-against). I understand that we want to highlight cases of direction discrimination, but the examples given make for quite harsh reading whereas the focus is how we're working towards better diversity, equality and inclusion. I personally think they could be removed and this definition would still have the desired effect.

guides/equality-diversity-and-inclusion/README.md
WillGibson: Done, hopefully I got it right ;-)

guides/equality-diversity-and-inclusion/policy.md
WillGibson: It's a valid question, but I think they should be here so that the flow of reading this document is not broken by having to go somewhere else.

guides/equality-diversity-and-inclusion/policy.md
StevenLeighton21, line 37: Suggest we say `rejecting a job applicant because of their religion or religious beliefs` rather than calling out specific examples.

WillGibson: I'm not convinced linking to an external definition is helpful, but I have changed the wording in the examples to make them less specific, which I think softens them a bit.

yaseminercan (approved):

Vikkigill (approved):

chadtomkiss-madetech (approved):

PR #416, opened 2020-11-12
Correct the information for pairers about when interview debriefs happen

WillGibson:


StevenLeighton21 (approved):
It does tend to vary (sometimes it's right after, sometimes it's 5pm etc.)

Could change to 'should be on the same day as the interview' but it's not a big distinction really

guides/hiring/pairing.md
```diff
@@ -12,7 +12,7 @@ Remotely, it's good to say hello to the other interviewers so they know you are
 
 You’ll need to be there for the pairing. This starts 30 minutes after the start of the interview. It lasts 30 minutes.
 
-You’ll need to be there for the ‘debrief’ to share your feedback, after the candidate has left. This is the last 30 minutes of the interview.
+You’ll need to be there for the ‘debrief’ to share your feedback. This will normally happen at 5pm on the day of the interview.
```
kjdchapman, line 15: Does the example under this need to change as well?

Are we actually doing debriefs this way right now? It's been a mix of both styles for me so far.

WillGibson: I have asked Michael McLean for clarification.

WillGibson: @StevenLeighton21 It might be worth removing your approval. I based the change on the reality I see on the ground this week, but if that's not a standard protocol the wording is not right yet.

WillGibson: Michael confirmed that the debriefing is at 5pm now. I've updated the wording based on what he sent.

PR #429, opened 2021-01-13
Update software_engineer_2.md

richmadetech:
Updated Engineer 2 Role description to be in line with other roles, and to correct the hiring statement that we are now hiring for this role.

### DIFF
--- a/roles/software_engineer_2.md
+++ b/roles/software_engineer_2.md
@@ -1,6 +1,6 @@
 # Software Engineer 2
 
-Bristol, Manchester & London, UK.
+Bristol, Swansea, Manchester & London, UK.
 
 Please apply for this role at [www.madetech.com/careers](https://www.madetech.com/careers).
 
@@ -10,7 +10,7 @@ Our Engineers deliver digital, data and technology outcomes that improve society
 
 At Made Tech we want to positively impact the future of the country by using technology to improve society. We help public sector organisations deliver quality software to help citizens get more from public services. To do this we work alongside brilliant public servants to modernise technology and accelerate digital delivery.
 
-As an Engineer 2 you will collaboratively lead planning of work and the development of features, coordinating with stakeholders and thier users, and seeking freedback and support when needed, to help influence the approach taken.  You will be comfortable managing complex changes within a single system, and seek support when managing change across multiple systems, ensuring that features meet project outcomes and user needs. 
+As an Engineer 2 you will actively contribute to planning of work and the development of features, helping to define the features and approach to delivering them. You will proactively pick up complex features, seeking guidance and support when needed to complete them, and be comfortable architecting and delivering simple features in new contexts. 
 
 You will have the opportunity to work across a variety of technologies and will need to be comfortable doing so. Our teams have used Ruby with Rails and Sinatra, ES6 with React and Angular, C# with .NET Core, Java with Spring, Python with Django, and plenty more stacks. We don’t limit ourselves as a company to particular technologies and we expect all our engineers to be keen on learning new skills.
 
@@ -75,7 +75,7 @@ Compensating you fairly:
 
 The salary for this role is location dependant:
 
-- Bristol: £36,900-£45,000
+- Bristol & Swansea: £36,900-£45,000
 - Manchester: £36,900-£45,000
 - London: £40,000-£50,000
 
--- a/roles/software_engineer_2.md
+++ b/roles/software_engineer_2.md
@@ -10,7 +10,7 @@ Our Engineers deliver digital, data and technology outcomes that improve society
 
 At Made Tech we want to positively impact the future of the country by using technology to improve society. We help public sector organisations deliver quality software to help citizens get more from public services. To do this we work alongside brilliant public servants to modernise technology and accelerate digital delivery.
 
-As an Engineer 2 you will collaboratively lead planning of work, seeking feedback and support when needed, influence the approach taken to deliver features for our customers and their users, and will be comfortable managing complex changes within a single system.
+As an Engineer 2 you will join a team delivering complex, modern solutions with a focus on meeting user needs. You'll be part of planning and completing work, seeking feedback and guidance when needed, and influence the approach and design of the systems you build.
 
 You will have the opportunity to work across a variety of technologies and will need to be comfortable doing so. Our teams have used Ruby with Rails and Sinatra, ES6 with React and Angular, C# with .NET Core, Java with Spring, Python with Django, and plenty more stacks. We don’t limit ourselves as a company to particular technologies and we expect all our engineers to be keen on learning new skills.
 
--- a/roles/software_engineer_2.md
+++ b/roles/software_engineer_2.md
@@ -2,17 +2,61 @@
 
 Bristol, Manchester & London, UK.
 
-Our Engineers build software that makes our customers happy, they prefer problem solving over task completing, love learning from colleagues, and constantly strive to be nice humans.
+Please apply for this role at [www.madetech.com/careers](https://www.madetech.com/careers).
+
+Our Engineers deliver digital, data and technology outcomes that improve society. They do this by delivering and architecting software.
 
 ## What does the job entail?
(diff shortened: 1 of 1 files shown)


roles/software_engineer_2.md
```diff
@@ -2,17 +2,61 @@
...
+
+Our Engineers deliver digital, data and technology outcomes that improve society. They do this by delivering and architecting software.
 
 ## What does the job entail?
 
-We primarily write and deliver custom software for the public sector. We work across central and local government, as well as in health, and our past lies in the technology startup world. Technical excellence for us isn’t about delivering to feature lists. We place a strong emphasis on outcome-based delivery; ensuring our customer’s goals are understood and achieved with the technology we deploy.
+At Made Tech we want to positively impact the future of the country by using technology to improve society. We help public sector organisations deliver quality software to help citizens get more from public services. To do this we work alongside brilliant public servants to modernise technology and accelerate digital delivery.
+
+As an Engineer 2 you will collaboratively lead planning of work, seeking feedback and support when needed, influence the approach taken to deliver features for our customers and their users, and will be comfortable managing complex changes within a single system.
```
StevenLeighton21, line 13: I'm not sure this directly lines up with Engineer 2 expectations, was it copied from a more senior role?

roles/software_engineer_2.md
```diff
@@ -2,17 +2,61 @@
...
+
+High performing software delivery teams need to be empowered to iteratively and rapidly deliver changes all the way through to production. To do this we combine our extensive cloud automation knowledge with DevOps culture.
+
+We've been using AWS from the start and as Advanced Partners are go to experts within the public sector. We use a range of IaaS, PaaS and FaaS depending on the needs of our users, in this case software teams, such as EC2, Lambda, ECS, Kubernetes, Heroku, CloudFoundry, Azure App Services, and more. We use VPC and PrivateLink for connecting to on-premise, legacy systems. We also use API Gateway, S3, CloudFront, SQS, SNS, SES, RDS, and many other services provided by AWS. You will have the opportunity to use and learn these technologies and we offer official AWS and other cloud training.
```
StevenLeighton21, line 19: This and the following two paragraphs are for a DevOps centric role - I don't think this is the case generally for Engineer 2, though might be relevant for roles we're currently hiring for?

roles/software_engineer_2.md
richmadetech: Yea happy for stuff to be pulled out and adapted - we are looking to hire Engineer 2s now, and I figured I'd expect them to have the same type of experience as the Academy - so I've softened some of the expectations, happy for others to be addressed and advice on this too.

roles/software_engineer_2.md
richmadetech: Great, be good to see what Hiring are looking for and to clarify - would Engineers from the academy be given some DevOps experience I guess?

roles/software_engineer_2.md
richmadetech: Oh, additionally my understanding is we are looking to hire on the higher end of Engineer 2, so would expect more experience, i.e. on track to Senior I think. But be great to get clarified

StevenLeighton21, line 13:
```suggestion
As an Engineer 2 you will join a team delivering complex, modern solutions with a focus on meeting user needs. You'll be part of planning and completing work, seeking feedback and guidance when needed, and influence the approach and design of the systems you build.
```

roles/software_engineer_2.md
StevenLeighton21: Almost certainly, on reflection I think it's actually the last paragraph that set off this comment, will comment there separately

roles/software_engineer_2.md
```diff
@@ -2,17 +2,61 @@
...
+
+We ensure we document our architecture and infrastructure as code, using technologies such as Terraform and OpenAPI. Containerisation is a big part of empowering our teams to develop, deploy and scale their applications, but so too is using AWS Lambda and avoiding the complexity of stateful services altogether. Right tool for the job.
+
+For us, DevOps is about culture rather than roles and titles. Even though this role is for someone with strong DevOps experience, the biggest impact you will have is coaching and helping teams use the platforms you build. You won't be building infrastructure in isolation or charged with deploying other peoples work into production. You'll empower teams with the mantra: you build it, you run it!
```
StevenLeighton21, line 23: I think this paragraph should go, we don't (I think) expect E2 hires to coach DevOps skills, and the 'Even though this role is for someone with strong DevOps experience' line is misleading, I think.

roles/software_engineer_2.md
```diff
@@ -2,17 +2,61 @@
...
 
-Our teams have used Ruby with Rails and Sinatra, ES6 with React and Angular 2, C# with .NET Core. We don’t limit ourselves as a company and we expect all our Engineers to be keen on learning new technologies. Automation is important to our teams, so we make sure there is a CD pipeline set up to build, test, and release. Also, we will usually be responsible for setting up a customer’s infrastructure too! Such as on AWS, GCP or Azure using a tool like Terraform, though sometimes we opt for a Platform as a Service like Heroku.
+We grow a team of polyglot programmers, which you might already consider yourself to be, who are versed in a mix of paradigms such as object-oriented, functional, declarative, event-based and aspect-oriented. You will need to be comfortable sharing your knowledge and skills with others, and be willing to learn new technologies and skills. 
 
 ## What experience are we looking for?
```
StevenLeighton21, line 27: Consider checking with the current brief for the E2 roles we're hiring for, I suspect some of these desirable/optional points are more applicable to more senior roles, for instance `Writing blogs and giving talks`

We love when our engineers do these things, it's something I wouldn't expect a less senior applicant to have done though!

richmadetech: It was in the SFIA examples, I did actually pull it out originally for exactly the reason you said, but dropped it back in since its part of SFIA and its an "optional" thing.

roles/software_engineer_2.md
StevenLeighton21: Sure thing - it's fine to tweak to reflect our current open positions, I think. Better to accurately reflect the E2 roles we're hiring for, see comment about desirable experience (maybe we keep the list longer in that case)

roles/software_engineer_2.md
richmadetech: Ah great spot, I think I missed this role is for... I'll bin the paragraph

StevenLeighton21: Given that, Not sure the preceding two paragraphs are ultra relevant, going into detail about our devops process? Unless you're hiring specifically in that space

roles/software_engineer_2.md
```diff
@@ -2,17 +2,46 @@
...
 
 ## What does the job entail?
 
-We primarily write and deliver custom software for the public sector. We work across central and local government, as well as in health, and our past lies in the technology startup world. Technical excellence for us isn’t about delivering to feature lists. We place a strong emphasis on outcome-based delivery; ensuring our customer’s goals are understood and achieved with the technology we deploy.
+At Made Tech we want to positively impact the future of the country by using technology to improve society. We help public sector organisations deliver quality software to help citizens get more from public services. To do this we work alongside brilliant public servants to modernise technology and accelerate digital delivery.
+
+As an Engineer 2 you will collaboratively lead planning of work and the development of features, coordinating with stakeholders and thier users, and seeking freedback and support when needed, to help influence the approach taken.  You will be comfortable managing complex changes within a single system, and seek support when managing change across multiple systems, ensuring that features meet project outcomes and user needs.
```
StevenLeighton21, line 13: I think cut this paragraph - I think 'collaboratively lead' is wording from the senior role, and I think the rest describes behaviours from SFIA 4 + that we wouldn't necessarily expect from an applicant. You could reword but I think it would read fine just cutting it.

richmadetech: Cool, reworded to soften the Lead part

StevenLeighton21 (approved):

PR #431, opened 2021-01-14
Steve/guidance performance

StevenLeighton21:
Add initial draft guidance around performance management.


StevenLeighton21:
Gonna go ahead and merge - was checked for process accuracy before raising the PR, and we can always review and change something that exists rather than coming back to a stale PR.

PR #436, opened 2021-01-22
Clarify use of adoption leave alongside maternity and paternity leave

ccp92:
Added some clarification about the use of adoption leave as we were previously only mentioning our offerings for maternity and paternity leave.


davidwinter:
@yaseminercan are we ok to merge this in now?

ccp92:
> @yaseminercan are we ok to merge this in now?

Apologies @davidwinter, Yas gave me a couple of clarifications to add and then it's good to go. Will try and get that done tomorrow 👍

PR #456, opened 2021-03-15
Move to build and test to github actions

jbevan4:
# What 
Removing a dependency on Circle CI to github actions for continuous integration

# Why
Removing an external dependency, such as CircleCI, allows us to keep everything within one service. Github actions have matured a lot and should be free for our use case. This means that users can stay within the "Github" sphere for all areas of the handbook, rather than needing look elsewhere for builds 👍


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,28 @@
+name: Pull Request Action with Caching
+
+on: pull_request
```
yusufsheiqh, line 3: Is there a reason to have this check only on a pull request?

yusufsheiqh (commented):
Do you think it would be better to have CircleCI and GitHub Actions working simultaneously? Once we confirm GitHub Actions is working as intended, we could remove CircleCI(?).

.github/workflows/pull_request.yml
jbevan4: I could do it on push instead. The idea was to keep the number of builds low to remain within free tier. I can see a situation though were you push to a branch which is already at the pull request stage. 

I think upon reflection, push sounds better?

MrTortoise: msg me when this is resolved to your liking and i can merge

MrTortoise (approved):

yusufsheiqh (approved):

jbevan4:
Ready to go @MrTortoise 👍

PR #458, opened 2021-03-22
cycle-to-work-scheme-limit. Removes outdated reference to £2.5k limit.

CorinWilkins:


benefits/cycle_to_work_scheme.md
```diff
@@ -2,7 +2,7 @@
 
 We take part in Cyclescheme as we wish to support those who wish to cycle to work whether it's for pleasure, aiming to be healthier, trying to be more green or any other reasons.
 
-Made Tech have a limit of £2500 with Cyclescheme which can be used to purchase a bike and any accessories you will need. You then pay a monthly 'salary sacrifice' to effectively 'hire' the bike and kit, over a 12 month period. Once this period has been completed you will be given options by Cyclescheme around owning your bike. If you leave Made Tech before paying off the purchase, the balance will be taken from your final salary payment.
+Made Tech use Cyclescheme, who will provide you with an voucher (with no price limit), which can be used to purchase a bike, parts and components, and any accessories you will need. You then pay a monthly 'salary sacrifice' to effectively 'hire' the bike and kit, over a 12 month period. Once this period has been completed you will be given options by Cyclescheme around owning your bike. If you leave Made Tech before paying off the purchase, the balance will be taken from your final salary payment.
```
bencurrerburgess, line 5: `an voucher` should be  `a voucher`

bencurrerburgess, line 5: I would remove the `(with no price limit)` if price limits change we shouldn't have to update this page, just let the reader refer the to Cyclescheme directly for Ts & Cs

baarkerlounger: "Options around owning" reads a little vague? Is there any more detail/guidelines on what those are likely to be?

leaky: The £2,500 limit was only added to the handbook recently. Has this been confirmed with ops?

CorinWilkins: @leaky I was hoping they would chime in here. In terms of the no price limit, I believe that is cyclescheme themselves. "Cyclescheme no longer has a limit! This means you can save money on everything you need. Go for a bike, a bike and accessories, or just accessories – it’s your choice!" Taken from their website.

CorinWilkins: @baarkerlounger if I follow Ben's advice below and leave the fine details to cyclescheme we should do the same here. However, it isn't the easiest thing to wrap one's head around.

baarkerlounger: Could maybe link the options directly (https://help.cyclescheme.co.uk/article/42-what-is-an-ownership-fee) here so that at least it's quick to look up? I actually hadn't realised the scheme itself was this complicated...

CorinWilkins: Done and done.

chrisblackburn: Thanks for this update. As @leaky suggests this was updated not too long ago from the previous Cyclescheme limit of £1,000 to a Made Tech limit of £2,500.

Up for other proposals - just a couple of considerations we should bake in here:

1) Made Tech is effectively providing these loans to employees. We'll want to maintain a manageable loan book. I doubt there'd be demand for 10 fully spec'd Dogma F12s in a month, but we'd want to avoid being open to that sort of outlay

2) More pertinently we need a hassle free way to recoup these loans from anybody who leaves - ideally as a single deduction from a final salary payment rather than needing a mechanism for managing and potentially chasing repayments from the leaver

PR #473, opened 2021-04-14
Create hiring_contractors.md

leeihparker:


danielburnley:
This will fail the build until we ignore the urls that are private and cause a failure:

```
 [dead(403)] https://docs.google.com/forms/d/...
 [dead(0)] https://made-tech.workable.com/backend
 ```

danielburnley:
Closed for #474 as it requires changes from there

PR #474, opened 2021-04-14
Add guidance around hiring contractors

danielburnley:
Brings in the change from #473 as well as ignores any links added that cause build failures


leeihparker (commented):
Tidied up line break on a couple of points

PR #485, opened 2021-05-19
Fix markdown rendering for some of the benefits files

LewisDaleUK:
The benefits box and flexible benefits markdown files were missing the .md file extension, so weren’t rendering properly when viewed in the browser


chubberlisk (approved):
🚢

PR #499, opened 2021-07-20
Add a new laptop security guide.

paul-mc:
Add a new guide detailing some Cyber Essentials Plus security requirements for laptop setup. Include some specific Linux (Ubuntu) config.  Also update the Linux section of the VPN guide.

This is a work in progress and should probably have Windows/MacOS specifics adding before merging.


paul-mc:
@baarkerlounger would you mind having a review of the Ubuntu bits?  Did you do anything else or anything different?

guides/vpn/README.md
```diff
@@ -37,7 +37,20 @@ That is it! You should now be connected to the `Made Tech - VPN`. If you are exp
...
 ### Ubuntu/Debian
 
-https://docs.foxpass.com/docs/foxpass-linux-l2tpipsec
+The Made Tech VPN uses Layer 2 Tunneling Protocol which is not installed by default, but can be installed using:
+
+```bash
+sudo apt install network-manager-l2tp network-manager-l2tp-gnome
+```
+
+Once that's done create the VPN connection by:
```
baarkerlounger, line 46: This looks roughly the same as what I did but for reference I didn't need to install anything all required libraries were already there (may have done so earlier though I suppose).

For gateway I used `madetechvpn.uksouth.cloudapp.azure.com`

In PPP Settings I had to check MSCHAP/MSCHAPv2 authentication options to be able to use MPPE

baarkerlounger (approved):

guides/it/laptop-security.md
```diff
@@ -0,0 +1,70 @@
...
+Your system should *not* auto mount or auto run files when media, such as a removable USB disk, is inserted.
+
+### Linux (Ubuntu)
+In Ubuntu this feature can be disabled by selecting _"Never prompt or start programs on media insertion"_ in _Settings_ > _Removable media_. 
+
+## VPN 
+You must be have [Made Tech's VPN](../vpn/README.md) configured on your system.
+
+## Anti-virus
+You must be running Anti-virus software with an up to date virus database.
```
baarkerlounger, line 67: Worth adding install instructions here as well I think. I ended up installing ClamAv:

```
sudo apt-get install clamav clamav-daemon clamtk


```

baarkerlounger: (tk for gui)

guides/it/laptop-security.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+## Account locking
+User accounts should be configured to lock after a maximum of 10 failed login attempts.
+
+### Linux (Ubuntu)
+To enable account locking on Ubuntu add the following line to `/etc/pam.d/common-auth`:
+
+```
+auth required pam_tally2.so onerr=fail deny=6 unlock_time=300 audit even_deny_root root_unlock_time=600
```
acodeninja, line 29: The documentation says 10 failed attempts above, but this config would lock after 6 (`deny=6`) this might cause some confusion.

paul-mc: yeah, good point - 10 is the maximum allowed, but that felt like a lot, so I opted for 6.  We probably need a Made Tech policy on this, but I'll clarify for now.

cob16 (approved):

guides/vpn/README.md
paul-mc: thanks @baarkerlounger - I've updated that section

guides/it/laptop-security.md
danielburnley: When we went through Cyber Essentials Plus last time - ClamAV was considered not good enough, as it doesn't _prevent_ you from opening an infected file - which is why we went with Sophos back then. Is this still the requirement/does clamav now support it?

(The test they did was giving an infected file, and then attempting to open it and if nothing stopped you it was considered a fail)

paul-mc: thanks @danielburnley that's a good question, I'll try to clarify internally and will update when I find out

baarkerlounger: For the test we just did it seems that as long as it doesn't auto execute on browser or email client download (i.e. it prompts you first) that seemed good enough.

paul-mc: This was an issue for Cyber Essentials Plus - I've updated the guide with
- some generic requirements for AV software 
- guidance for installing ClamAV and enabling on-access scanning - which hopefully does meet the requirements.

PR #502, opened 2021-07-28
List config instructions explicitly since Medium article is behind a log in wall

baarkerlounger:
…and may not be accessible forever.


guides/it/laptop-security.md
```diff
@@ -75,10 +75,119 @@ Made Tech do not have a preferred/approved Anti-virus tool for Linux, but it is
 #### ClamAV
 [ClamAV](https://www.clamav.net/) is an open source AV tool that is popular on Linux - it can be configured in a way that meets the requirements above, but this involves enabling on-access scanning.
 
+In can be installed from the Ubuntu repositories:
```
paul-mc, line 78: I'd deliberately not included instructions, partly because I didn't want us to have to maintain them, but also because they're quite Ubuntu specific and I suspect there are other user distros in use.  But on reflection I think it's a good idea to include them.

Do you think it's worth breaking the AV section into a separate guide, like the the VPN guide?

build.js
```diff
@@ -17,12 +17,13 @@ function checkFile (fileName) {
...
 
       const baseUrl = `file://${path.dirname(path.resolve(fileName))}`
-      
+
       const ignorePatterns = [
         { pattern: /www.glassdoor.co.uk/ },     // glassdoor returns 503 status to circle ci hosts
         { pattern: /www.aws.training/ },
         { pattern: /www.certmetrics.com/ },
         { pattern: /made-tech.workable.com/ },
+        { pattern: /clamav.net/ },
```
paul-mc, line 26: thanks, I was going to look into how to fix this :+1:

guides/it/laptop-security.md
```diff
@@ -75,10 +75,119 @@ Made Tech do not have a preferred/approved Anti-virus tool for Linux, but it is
...
+
 The following resources are useful for configuring ClamAV:
 - The [Configuration](https://docs.clamav.net/manual/Usage/Configuration.html) section of the ClamAV documentation
 - The [On Access Scanning](https://docs.clamav.net/manual/OnAccess.html) section of the official docs
-- [Installation & Configuration of ClamAV Antivirus on Ubuntu 18.04](https://aaronbrighton.medium.com/installation-configuration-of-clamav-antivirus-on-ubuntu-18-04-a6416bab3b41) - this Medium article was 
+- [Installation & Configuration of ClamAV Antivirus on Ubuntu 18.04](https://aaronbrighton.medium.com/installation-configuration-of-clamav-antivirus-on-ubuntu-18-04-a6416bab3b41) - this Medium article was
 particularly helpful for installing on Ubuntu and enabling on-access scanning.
-
-The official Ubuntu PPA does not always provide the most recent version - at the time of writing the official PPA ships clamAV `0.103.2`, the most recent version is `0.103.3`.  Given that `0.103.3` was release more than 30 days ago, the PPA does not meet the requirement for running the latest version.
```
paul-mc, line 84: I think we should leave this is following the conversation with Sam

baarkerlounger: Was just about to amend it to include compilation instructions. Will split this out into a separate section at the same time.

paul-mc (approved):
Looks good to me now @baarkerlounger :+1:

PR #507, opened 2021-08-20
Adding holiday pro-rata time to guidance

yaseminercan:


jessicalondon:
Interesting change! Won't the implementation mean employees will take less holiday? 

In a scenario where someone was to take holiday towards the start of the year and leave towards the end, they would be financially punished for doing so - Which feels weird alongside the unlimited holiday benefit. 

Just my 2cents. Keep up the great work!

leaky:
> Interesting change! Won't the implementation mean employees will take less holiday?
> 
> In a scenario where someone was to take holiday towards the start of the year and leave towards the end, they would be financially punished for doing so - Which feels weird alongside the unlimited holiday benefit.
> 
> Just my 2cents. Keep up the great work!

@jessicalondon this addition to the handbook is reflective of what's in our employee contracts and so employees will already be already be aware that any days accumulated in excess of the pro-rata'd _contracted_ allowance will be recouped. A contracted 23 days is a little below average however and I would like to see this number increased to what seems like an industry-standard of 25 days in the future which will result in less-harsh financial penalties for those employees that do decide to leave part-way through the year.

PR #529, opened 2021-11-05
D&I minutes update

ccp92:
Following a discussion about people being hampered in editing the D&I meeting minutes via GitHub, it was agreed that going forwards, we would move these notes into a publically viewable Google Doc to improve the accessibility of the minutes.


lukemorton (approved):
LGTM :)

PR #557, opened 2022-01-21
Add link to Workable Referrals on the referral policy page

sbt92:


adrianclay (approved):
Great idea adding the link, definitely an improvement.

Minor thought, I wonder if adding the link to the first sentence would improve the flow, and reduce the overall amount of text to read.  Something along the lines of...

> feel free to [refer them via Workable]

sbt92:
@adrianclay Thanks for your input that seems like a more logical place. If you could give me another review that would be appreciated.

WillGibson (approved):

PR #559, opened 2022-01-25
Update link check

cob16:


cob16:
The link looks to be broken for real, but that proves the update is working as intended. Merging anyway as the dep update is important

PR #567, opened 2022-02-21
Add a link to a new pairing guide for Engineers

irenegrandeh:
Added a link to a guide that explains in detail what an Engineer needs to do if they are interviewing a candidate.


sbt92 (approved):
LGTM - re the CI failure the dead link is a false positive

PR #570, opened 2022-03-02
Fix brand spelling, GoPro

NyWeb:
This is of course a small change, but it is to help Nina and Allison understand GitHub


sbt92 (approved):
LGTM

PR #578, opened 2022-04-04
Updating DM Roles

lburnett88:
- Adding in all Delivery Job Titles
- Linking all roles to the Delivery Manager Role until individualised Job Descriptions are agreed
- Updated SFIA Levels 
- Removed links to bespoke SFIA Levels


yusufsheiqh (commented):
From an external perspective, it might be better for the roles without a role description to not have the link. 

Links can be added in as the pages get added, and this would be applicable for the SFIA pages as well.

yusufsheiqh (approved):

PR #587, opened 2022-04-14
Principles

mackella:
### What

Added a set of guiding principles which describe how Made Tech believe delivery teams should behave and work with clients.

These principles were created collaboratively by a group representing capability practices and the delivery management organisation. They're published in a draft state initially and feedback is welcomed, either via Github issues or in #principles-feedback in Slack.

### Why

There are regular requests (from new starters particularly) asking for better guidance on how Made Tech people should  approach client deliveries. 

These principles are the first step towards defining a holistic range of guidance which derive from our core values and culminates in specific, practical recommendations on tools and practices. We will call this the Made Tech Way.

### DIFF
--- a/team-norms/README.md
+++ b/team-norms/README.md
@@ -10,7 +10,7 @@ Our accepted practices for structuring deliveries.
 
 Our expected reporting standards for delivery teams.
 
-## [The Made Tech Way](principles.md)
+## [The Made Tech Way (DRAFT)](principles.md)
 
 Our principles describing how we deliver in teams and work with clients.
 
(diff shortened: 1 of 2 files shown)


mackella:
> If we're being true to leaving things better than we found them, could this `team-norms` folder benefit from a further tidyup from a relevance/how we actually currently work perspective as part of this change?

Yes, definitely. All of those pages should be replaced ultimately by the practices content that will sit under these principles. It's tempting to just remove everything under `team-norms`, but there is some still relevant content in there. I'm happy to have a go at tidying it and I also want to fix the broken links that are failing the build currently 😬.

(Updating some of that content is on the new chalet backlog already: https://trello.com/c/NGaoKuQ1/32-review-and-improve-the-development-practices-section-in-the-handbook)

lukemorton (approved):
LGTM

HarryTrimble (approved):
Looks good. Nice adjustments @jameshollo

PR #590, opened 2022-04-20
Update pairing.md

wystanpalm-madetech:
Modifying this guidance to require that Senior or Lead Engineer candidates choose one of Tennis or Bowling katas as the others may not have enough complexity to be able to assess the candidate at that level.

### DIFF
--- a/guides/hiring/pairing.md
+++ b/guides/hiring/pairing.md
@@ -51,7 +51,7 @@ Tell the candidate the key facts:
 - Ask if they’re familiar with pair programming. If not, explain briefly.
 - Decide on a pairing style together.
 - Ask if they’ve done a kata before. If not, explain briefly.
-- Ask which kata they’d like to do. If they’re not sure, suggest one from [this list](https://learn.madetech.com/katas/). Tennis or Bowling work well, and if the candidate is applying for a Senior or Lead Engineer role, then they must pick one of these two.
+- Ask which kata they’d like to do. If they’re not sure, suggest one from [this list](https://learn.madetech.com/katas/). Tennis or Bowling work well and, if the candidate is applying for a Senior or Lead Engineer role, then they must pick one of these two.
 - Ask if they’re familiar with test-driven development (TDD). If not, explain briefly.
 - Tell the candidate you’d like to write tests as part of doing the kata.
 - If you’re remote, decide how you’ll pair. You could use Tuple, Live Share, or screen sharing.


adrianclay:
> may not have enough complexity to be able to assess the candidate at that level

Which of the "What you are looking for" necessitates complexity?

CorinWilkins:
While I agree with Adrian, I think it is an issue if a candidate can complete a kata in less than 15 minutes. If we don't want to restrict certain katas to levels then maybe we should not use the easier ones for interviews.

adrianclay:
> While I agree with Adrian, I think it is an issue if a candidate can complete a kata in less than 15 minutes. If we don't want to restrict certain katas to levels then maybe we should not use the easier ones for interviews.

Agreed that if the problem we've found is individuals racing through a kata, I don't see why the guidance should be "do one of these katas if you're applying for senior or lead".  I don't think the SFIA level or kata choice would change somebody's ability to revise a particular kata, and type out a prepared solution.

reub-the-cube:
I've not been an interviewer at Made Tech (yet!) but if the experience of the interviewers is that candidates are completing the katas too quickly, perhaps our messaging could be looked at? The exercise isn't about completing the problem, but about how you work towards the solution. I was comfortable coming in that I wasn't going to get near completing the kata but I had experience on the other side of the hiring process at my previous role. Also, just to throw it out there, I actually think bowling is one of the simpler ones :) 

> Modifying this guidance to require that Senior or Lead Engineer candidates choose one of Tennis or Bowling katas as the others may not have enough complexity to be able to assess the candidate at that level.

I feel there's enough complexity in most katas: for example the video game store, mars rover, numeral to lcd with extension should all have some level of flexibility in them and the candidate should be able to react to change or take feedback on board if they manage to steam through the exercise.

> I think it is an issue if a candidate can complete a kata in less than 15 minutes

Completing the kata too quickly would hopefully mean that the candidate was diligent and well-practiced but you would also hope they chose one that would fill the time otherwise they won't have the opportunity to showcase their skills. As above, I think the time can be filled by probing and perhaps making it clearer to the candidate.

> I always say to the candidate "the pairing isn't a coding test", but if we are judging their ability to solve a complex problem, then perhaps I shouldn't be saying that.

In my previous role, I have always said this too. For me it's always been about the approach to problem-solving, collaboration, TDD etc. Hence my suggestion above about giving the candidate something to react to or think about if they complete it quickly.

CorinWilkins:
I've raised another PR to remove Mumbling from the katas list, replaced with something a little more complex. It's a happier medium than dictating what kata a particular level engineer can do. Suggest we close this PR for the time being.

PR #606, opened 2022-05-12
Added the first draft of the Azure Certification page to the cloud gu…

petecraven91:
…ides for linking microsoft learning accounts to the MT azure partner portal


reuben-x-lab:
What do you think about putting these files under the (brand-new) [cloud-and-engineering](https://github.com/madetech/handbook/tree/main/communities-of-practice/cloud-and-engineering) page?

reub-the-cube:
What do you think about putting these files under the (brand-new) [cloud-and-engineering](https://github.com/madetech/handbook/tree/main/communities-of-practice/cloud-and-engineering) page?

petecraven91:
Didn't know that existed tbh. I think that is a wider question of what should live under there but this also applies to the data cop and the cyber cop

PR #608, opened 2022-05-12
Tidies up and modernises guidance for holiday bookings

chrisblackburn:
- Tidies up some ambiguous language around periods less than one week
- Updates to reflect current organisation structure, particularly the introduction of Industries and Capabilities
- Uses client instead of customer
- Makes more explicit the approval/agreement step


lukemorton (approved):
LGTM

PR #620, opened 2022-05-25
Update principles.md

jameshollo:
An iteration based on feedback gathered by Harry, and with another content pass from me and Lucy.


jameshollo:
"We recognise there's always a different way to solve a problem, often supported by proven practices and technologies."

At face value, this seems contradictory. "We like to work in a new and different way, by working in the old same ways." Do we want to revisit – or can we expand on it to solve for this?

PR #627, opened 2022-06-08
Incorrect unified memory

Porim:
MacBook Pro comes as standard with 16GB of memory and it can be upgraded to more, but I believe that is not the case since the Mac that I've received from Systemagic has 16GB.


cob16:
You are right, its an upgraded model that has the 32GB https://support.apple.com/kb/SP854?locale=en_GB

cob16 (approved):

PR #660, opened 2022-07-14
Changed CharlieHR references to HiBob

JazJax:


guides/learning/booking_learning_time.md
```diff
@@ -25,19 +25,19 @@ We want to ensure teams own their learning time and that includes thinking about
...
 
 ## How do I record this time?
 
 Just like booking a holiday
 
-Log into CharlieHR
+Log into HiBob
 
-Navigate to book time off, (once it’s been okayed  by everyone) and then choose Learning Time from the drop down menu.
+Click Request Time Off (on the home page) or My Time Off (on the Time page), (once it’s been okayed  by everyone) and then choose Learning Time from the drop down menu.
```
luke-a-williams, line 38: The 2nd bit is nearly there on the journey, technically it's also still "Request time off" from the time tab once you've selected "my time off". Both of those routes then bring you to the same screen - can you add that additional step in your wording?

agnieszkazbikowska-madetech (approved):
Looks good!

luke-a-williams (changes requested):
7/8 are perfect, and the last one I think is nearly there but needs to be a little more precise to unite the end point of the 2 journeys in - what do you think?

luke-a-williams (approved):
LGTM

PR #662, opened 2022-07-20
Single source of truth for timesheeting and updated guidance re 5 hour rule

WillGibson:
Why?

The Cloud and Engineering CARE group conducted a survey of engineers and found that many did not know about and or understand the "five hour rule".

This pull request seeks to better explain the rule so we can use it to help everyone understand how to apply it.

It also removes a second page of timesheet guidance we discovered during our conversation so that there is a single source of truth.


guides/process/scheduling/how_to_timesheet.md
```diff
@@ -40,6 +40,12 @@ You may be asked to support internal Made Tech initiatives while on a billed eng
 
 Where these commitments total less than 5 hours per week, they should be included in our standard operating overhead and should not be accounted for separately from billed hours.
 
+Note that some internal activities do not count towards this 5 hours as our clients are perceived to benefit from them indirectly. These include:
+
+* Hiring
+* Line management (1:1s etc.)
+* Taking part in communities of practice
```
WillGibson, line 47: Any more we should add here?

guides/process/scheduling/how_to_timesheet.md
```diff
@@ -40,6 +40,12 @@ You may be asked to support internal Made Tech initiatives while on a billed eng
...
+
 When the commitment exceeds 5 hours in a week, each element should be allocated individually to the appropriate project in Kimble and not included in billable hours.
```
WillGibson, line 49: It is not clear here whether we start allocating elements above 5 hours, or the whole lot.

If we do 7 hours, to we allocate 7 hours to non client categories or 2 hours?

guides/process/scheduling/how_to_timesheet.md
mackella: I think this is a bit confusing. Hiring *is* included for the purposes of the 5 hour break point, as explicitly stated in the first line of this section. Line management and CoP involvement (including CARE) are not.

guides/process/scheduling/how_to_timesheet.md
```diff
@@ -40,6 +40,12 @@ You may be asked to support internal Made Tech initiatives while on a billed eng
 
 Where these commitments total less than 5 hours per week, they should be included in our standard operating overhead and should not be accounted for separately from billed hours.
 
+Note that some internal activities do not count towards this 5 hours as our clients are perceived to benefit from them indirectly. These include:
```
adrianclay, line 43:
```suggestion
Some internal activities do not count towards the 5 hours as our clients benefit from them indirectly. These include:
```

guides/process/scheduling/how_to_timesheet.md
adrianclay:
```suggestion
When the total internal commitment exceeds 5 hours in a week, each commitment should be allocated in its entirety to the appropriate project in Kimble and not included in billable hours.
```

guides/process/scheduling/how_to_timesheet.md
WillGibson: Ah, I obviously got the wrong end of the stick from our meeting. kind of proves the value of this exercise! Will remove hiring.

adrianclay (approved):
Champion!

guides/process/scheduling/how_to_timesheet.md
WillGibson: How about this?
```suggestion
When the commitment exceeds 5 hours in a week the 5 hour rule no longer applies, so each element should be allocated individually to the appropriate project in Kimble and not included in billable hours.
```

WillGibson: Mashup of mine and @adrianclay's...
```suggestion
When the total internal commitment exceeds 5 hours in a week the 5 hour rule no longer applies, so each commitment should be allocated in its entirety to the appropriate project in Kimble and not included in billable hours.
```

WillGibson: I'm going with my mashup.

PR #679, opened 2022-08-18
Include info about exam expenses

st3v3nhunt:
Added section about how to expense an exam and the best way to do it given the AWS exams are charged in USD.


st3v3nhunt:
The content has been reviewed (and revised) by @VincentSmith93

PR #685, opened 2022-08-30
Add guidance on Jury service longer than 2 weeks question

WillGibson:


guides/jury_service.md
```diff
@@ -6,3 +6,7 @@ If you are summoned for Jury Service, the process is simple...
 * Book the appropriate amount of time off as...
   * Policy Type: Out of Office (Paid)
   * Reason: Jury Duty
+
+While at court...
+
+* If you are asked whether a trial lasting longer than 2 weeks will cause problems with your job, the answer is yes.
```
WillGibson, line 12: @yaseminercan Does this apply to all roles or just client facing roles?

yaseminercan: For consistency I would say all roles. It's probably best to check at the time with your Line Manager who can advise with context available closer to the event. For client facing roles I think the account team would be the check point. But generally, more than 2 weeks for Jury service is probably going to be disruptive for any team. (as we also need to ensure that sufficient holiday is taken by the team member)

WillGibson: How about this...
```suggestion
* If you are asked whether a trial lasting longer than 2 weeks will cause problems with your job, the answer is yes. If you have any questions, please check with your line manager and/or account team.
```

yaseminercan (approved):

PR #689, opened 2022-09-14
Update _delivery_manager.md

emmawebb-mt:
Created sub-heading for Risk & Quality to make this job consistent with other Delivery roles


lburnett88 (approved):
Approved, thanks for making this consistent

PR #695, opened 2022-09-21
Update Leave and Time Off

Camsmitti:


owenblacker:
@Camsmitti It's generally better to avoid using "here" as the text of a link. The link text is the thing that catches our eye and can be read out differently by screenreaders, for example, so making link text be descriptive is a big plus for accessibility.

Would you mind changing to from `for more information see our something policy [here](link)` to `for more information [see our something policy](link)` in each case, please? 💖

owenblacker:
Oh, also, could you change the filename from `guides/welfare/Leave and Time Off` to `guides/welfare/leave-and-time-off.md`, to match the other files and so that GitHub knows how to format it correctly?

If you need a hand, shout on Slack and I or someone else should be able to help 🙂

Thanks! 💖

PR #716, opened 2022-11-07
Update laptop_replacements.md

jamesLemontatmt:


rachsheppard (commented):
Can you add the <pre> </pre> back in on the address?
Also worth sense checking that the bullets are aligned as they're meant to be

rachsheppard (changes requested):
Sorry James. The test failure is because the finance@madetech.com link isn't valid - it needs the emailto part in front (you'll need to check exact syntax)

PR #721, opened 2022-11-10
Update software_licenses.md

jamesLemontatmt:


rachsheppard (commented):
Can we change remove Atlassian and add Trello.
Remove Reward Gateway as it's not really a tool, same with Torii 
Remove Zoom and Mural
Update Adobe to be 'Adobe Creative Cloud or relevant applications'
Remove Word and Excel

PR #738, opened 2022-11-21
Update pension_scheme.md

Camsmitti:
Added information for first 90 days deferred and max value as being 4% during this time.


AlliBenfield (changes requested):
The 4% is the employer contribution so needs clarifying a bit further.

Opt in/out and contribution changes now via bob rather than email so needs bob process adding.

NinaPinPeople (changes requested):
Please can you add the relevant info where I have left xxxx's @Camsmitti and then I'm comfortable for this to go live, thanks.

AlliBenfield (approved):

PR #750, opened 2022-12-15
Changing Flexible Leave to Fixed Leave

yaseminercan:
Flexible leave will end on 31/12/200
This update is to update the guidance to align with our updated holiday benefit which will make a fixed pot of leave available to all team members. The total holiday available is set to 30 days between June and May, changing from January to December, which is to align it to our financial year.

### DIFF
--- a/benefits/flexible_holiday.md
+++ b/benefits/flexible_holiday.md
@@ -48,7 +48,7 @@ It is the responsibility of the person agreeing to a holiday booking to ensure i
 
 * If a team members performance declines significantly due to abuse of this policy, then we reserve the right to review their use of this policy.
 
-* If you leave Made Tech part way through the year we will use 30 days to pro-rata your final salary. In some cases that means we'll pay you for accrued and unused holiday and in others it means you'll have to pay days taken over your pro-rated allowance back.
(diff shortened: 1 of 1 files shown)


benefits/flexible_holiday.md
```diff
@@ -1,12 +1,12 @@
...
 
-We have adopted a flexible holiday policy, where each team member is given the flexibility to take whatever holiday they need. We believe this policy will help us to perform at the highest possible level. We ensure everyone in the team take at least the minimum 23 days though welcome everyone to take more than this.
+All team members have 30 days of leave available between June and May (pro-rata for part time colleagues), plus bank holidays. We believe this policy will help us to perform at the highest possible level.
 
-It is important that you and your colleagues feel that holiday is highly available. That holiday is an easily accessible commodity that enables everyone to get the necessary rest that they need. We intend untracked holiday to prevent burn-out in situations, where individuals feel that they need to "save up" holiday days, or plan elaborate strategies around using your holiday allowance "effectively" later in the year, rather than taking holiday when needed most.
```
cob16, line 7: Would be nice to address the burn-out point in the new copy.

cob16 (approved):

benefits/flexible_holiday.md
yaseminercan: Agreed. We will make a further update for that. Just want to make sure the benefit is accurate right away, then we will come back for more edits :)

PR #806, opened 2023-01-05
Delete paid_anniversary_break.md

kimkaveh1:
Not sure I've done this correctly, but we need to delete this whole page. I've created a G doc with the content if we ever need to use it again in future.


Camsmitti:
Page is already deleted.

PR #863, opened 2023-04-13
Mark start00 linux set up

markStart00:
I cross-checked the handbook linux_av with the knowledge base (kb) [linux requirements](https://docs.google.com/document/d/13_Je82eCpPU2qZgYINxDYX9OxRrVemdixj8yD4IFH8s/edit).

The important sections I added:
- Full disk encryption
- Automatic updates enabled
- DriveStrike

I added a few helpful section and information taken from the kb linux requirements :
- Approved software
- Account locking details (I think these are iso requirments to pass the audit)
- ISO hyperlink ( it points to the Made Tech blog for iso 27001 - I thought it was helpful - but it can point to iso.org).

I ran a spellcheck (making it seem like I've made more changes than I have!):
- SentinalOne <-- SentinelOne
- license <-- licence (english spelling not sure if it was important to change).

And made some small changes welcoming Linux starters (although please delete it it isn't the right tone) and some small changes that felt more correct when I fresh installed my system to encrypt the disk and went through the setup check list again.
I think it will save time for a new starter and adds some helpful information from my new starter perspective! 
But please feel free to change anything. Or maybe adding a section to point new starters to the kb instead of the av ?

### DIFF
--- a/guides/it/linux_av.md
+++ b/guides/it/linux_av.md
@@ -61,12 +61,12 @@ This will:
 
 ### Linux (Ubuntu) 22.04+
 pam_tally2.so has been replaced by pam_faillock.so <br>
-pam_failock.so settings can be made in `/etc/security/faillock.conf` and by adding the following to `/etc/pam.d/common-auth`:
+pam_failock.so settings can be made in `/etc/security/faillock.conf` and by adding the following to the `/etc/pam.d/common-auth` file:
 ```
 auth    [default=die]   pam_faillock.so     authfail
 auth    sufficient      pam_faillock.so     authsucc
 ```
-If you have prevented all logins by mistake with pam_tally2 you can reboot to recovery mode in# All about Linux
+If you have prevented all logins by mistake with pam_tally2 you can reboot to recovery mode into a root shell to access `/etc/pam.d/common-auth`.
 
 #### Root account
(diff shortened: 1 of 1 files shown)


awilkins:
This seems fine ... has drawn my attention to some omissions in this document as well (if it's going to be a component of the source of truth for Linux setup).

markStart00:
The public facing linux_av and #linuination genuinely got me excited about joining Made Tech.

But it may be better to point to the internal facing knowledge base for the actual steps.
( I think they are the more formal documentation of the process for the iso audits ). 
I can add a section - something like - 'After joining ... the kb ... for the more detailed formal process'
but I think it is a good idea for the linux_av to be the source of truth for the setup.
It is what most new starters will see first.
From my new starter perspective I didn't find out about the kb, until after I had set up my system.

Are you sure about my positive tone, I like it, I wanted to try to add to the vibe that made me interested.  
I guess it's a small change and can always be adjusted in the future.

awilkins:
Yeah, I have no quibbles with positive tone - I genuinely think increasing the Linux engineer count at Made will be beneficial, at the moment this is a small fraction of the engineers, I suspect the fastest path is by conversion rather than recruitment, but every little helps :-)

PR #881, opened 2023-05-10
Update ethical_boundaries.md

NinaPinPeople:
Change to reflect client language and reference to project assignment guidance.


NinaPinPeople:
@sampaice-MT be good to get another set of eyes on this before it's published please?

PR #933, opened 2024-02-12
Update Hardware.md

richrace:
There is no M3 Pro with 8 Core CPU / I received a M3 8 Core CPU


rachsheppard:
Typo on my part on the page. Rich's correct is correct.

rachsheppard (approved):
Typo on my part on the page. Rich's correction is correct.

PR #946, opened 2024-07-03
Update and rename principal_ucd.md to principal_ucp.md

olimonks-mt:
Update to the JD for our UCP Principal role


NinaPinPeople (approved):
Looks good @olimonks-mt

PR #980, opened 2025-03-27
Clarify how pension contributions can be adjusted

adrianclay:
Remove sentence which said that this can only happen once a year.

Add content about maximum contributions not being higher than minimum wage.

### DIFF
--- a/benefits/pension_scheme.md
+++ b/benefits/pension_scheme.md
@@ -9,7 +9,9 @@ The pension contribution available to you is aligned to your SFIA level:
 - SFIA 6: 8% employer matched
 - SFIA 7: 9% employer matched
 
-The minimum contribution in every level is 4% (employer matched), there is no maximum employee contribution at present.
+The minimum contribution in every level is 4% (employer matched).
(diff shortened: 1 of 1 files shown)


benefits/pension_scheme.md
```diff
@@ -9,15 +9,16 @@ The pension contribution available to you is aligned to your SFIA level:
 - SFIA 6: 8% employer matched
 - SFIA 7: 9% employer matched
 
-The minimum contribution in every level is 4% (employer matched), there is no maximum employee contribution at present.
+The minimum contribution in every level is 4% (employer matched).
+There is no maximum contribution, however your salary must remain above the national minimum wage thresholds set by HMRC.
```
WillGibson, line 13: What happens to pension contributions if you earn less than minimum wage?

Is earning less than minimum wage even a concern at Made Tech? I would hope not.

NinaPinPeople: @WillGibson whilst we pay above the NMW by a chunk, if someone were to make significant changes to their employee contributions via salary sacrifice, in essence HMRC will only look at the remaining salary earnings to consider NMW so it can become an inadvertent issue we have to consider and give people a heads up on.

PR #1036, opened 2025-04-15
Make link checking GitHub action work again

WillGibson:
The check has been failing since November :-|

PR #1038, opened 2025-04-25
Update work_ready.md to clarify laptop wording

TimHacker:
It would make sense to clarify that the PC is optional rather than being standard for engineers as this was a question an external candidate asked recently. They were keen on the Macbook!


WillGibson (approved):

WillGibson:
https://github.com/madetech/handbook/pull/1036 will fix the failing link check thing.

